Issue #20, opened 2019-03-05
As a developer I want to have a Code of Conduct to abide by


marickmanrho:
As suggested by Robin, It is probably good to have a Code of Conduct. [https://www.thebalancecareers.com/code-of-conduct-1918088]()

PR #21, opened 2019-03-05
#20 Include Code_Of_Conduct

marickmanrho:
According to Contributor-Convenant


Drvanon:
Excellent, @ddvduuren and I had been talking about this, but your suggestion is very welcome.

PR #30, opened 2019-10-02
Mock data

PJvanVliet:
Changed MovementsPage and TimelinePage to accept and display mocked data.


gridt/src/api/model/userLastUpdate.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+/**
+ * This is the last update that the user has given in this movement.
+ */
+export interface UserLastUpdate { 
+    /**
+     * Date time of the last update of the user.
+     */
+    timeStamp?: string;
```
marickmanrho: The timeStamp being a string... Not checking if it is actually a timestamp might be problematic in the future.

gridt/src/api/api/default.service.ts
```diff
@@ -0,0 +1,539 @@
+/**
```
marickmanrho: I'm just going to assume this is alright.

PR #48, opened 2020-03-27
Issue 47

AndraUdrescu:
See if everything is fine


Drvanon:
You are not able to check if everything works right now, as the api URL is still set to http instead of https. Which brings me to another point: there is a good chance that you will need to do some refactoring in the api as well, to make sure that the email is stored there as well.

For now I understand that you leave the Email/username at the login, but in reality we will work with only emails in the future, so I recommend that you choose to leave either email or username but not both.

Drvanon:
Here is a link to the code in the api where username is being stored:
https://github.com/GridtNetwork/gridt-client/blob/a1d7bd0bf31a7231b5158357b7684a68001c3d92/gridt/src/app/api/api.service.ts#L94

Drvanon:
Very minor thing: in your request a change has been mode to package-lock.json. This file does not belong in helping solve this specific issue and should therefore not be included.

gridt/src/app/login/login.page.html
```diff
@@ -27,7 +27,7 @@
               lines="none"
             >
               <ion-label>
-                Should be a valid email address.
```
Drvanon: Because we have decided for an email address only login, you can leave email here.

gridt/src/app/login/register/register.page.html
```diff
@@ -11,6 +11,24 @@
...
+                type="username"
+                ngModel
+                name="username"
+                required
+                #usernameCtrl="ngModel"
+              ></ion-input>
+            </ion-item>
+            <ion-item
+            *ngIf="!usernameCtrl.valid && usernameCtrl.touched"
+            lines="none"
```
Drvanon: Can you indent these lines for clarity?

Drvanon:
In hind sight, that commit did not belong in this branch. Oh well.

AndraUdrescu:
The package was changed due to update so I have the most updated version now.  Made the changes as well

PR #57, opened 2020-04-10
Android build

AndraUdrescu:
Mostly android file changes, had to do those so I could build. Also there is a build

### DIFF
--- a/gridt/android/app/src/main/AndroidManifest.xml
+++ b/gridt/android/app/src/main/AndroidManifest.xml
@@ -8,7 +8,8 @@
         android:label="@string/app_name"
         android:roundIcon="@mipmap/ic_launcher_round"
         android:supportsRtl="true"
-        android:theme="@style/AppTheme">
+        android:theme="@style/AppTheme"
+    >
 
         <activity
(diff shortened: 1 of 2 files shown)


Drvanon:
I think we are going to go for PWA mostly now, so the underlying base (android) has become redundant. 

For future reference: does this still only built with that added line, or will it also work without it?

AndraUdrescu:
The android thing I am pretty sure it's an added package. As long as you remove that package the android stuff should be gone.

    On Saturday, April 18, 2020, 3:40:57 PM GMT+2, Robin Alexander Dorstijn <notifications@github.com> wrote:  
 
 


I think we are going to go for PWA mostly now, so the underlying base (android) has become redundant.

For future reference: does this still only built with that added line, or will it also work without it?

AndraUdrescu:
Most of the changes were inside Android specific files which you wouldn't need technically now. Personally I am going to make a build again mostly for testing for me cause it would be easier to have gridt as an app rather than use the side. And it would look nicer

Drvanon:
We can't remove the package as it forms an integral part of the authentication procedure. Let's leave it for right now and continue later.

PR #66, opened 2020-04-18
Make movement parts full screen on small screens on home page.

### LINKED ISSUE #60: The width of cards on the home page should match the width of the screen for small enough screens.
When there is not enough content to fill up the card, it becomes very tiny. It is good to have sensible min and max widths set for both the card and its lists.

marickmanrho:
Fixes #60

### DIFF
--- a/gridt/src/app/home/home.page.scss
+++ b/gridt/src/app/home/home.page.scss
@@ -6,3 +6,13 @@
     display: flex;
     justify-content: space-between;
 }
+
+// The width of cards on the home page should match the width of the screen for small enough screens. Issue #60
+@media screen and (max-width: 600px) {
+  ion-card{
+    display: flex;
+    flex-direction: column;
+    width: 100% !important;
+    margin: 0 !important;
+  }
+}
--- a/gridt/src/app/home/home.page.scss
+++ b/gridt/src/app/home/home.page.scss
@@ -7,9 +7,13 @@
     justify-content: space-between;
 }
 
+.movement-card{
+  min-width: 300px;
+}
+
 // The width of cards on the home page should match the width of the screen for small enough screens. Issue #60
-@media screen and (max-width: 600px) {
-  ion-card{
+@media screen and (max-width: 576px) {
+  .movement-card{
     display: flex;
     flex-direction: column;
     width: 100% !important;
--- a/gridt/src/app/home/home.page.scss
+++ b/gridt/src/app/home/home.page.scss
@@ -14,9 +14,7 @@
 // The width of cards on the home page should match the width of the screen for small enough screens. Issue #60
 @media screen and (max-width: 576px) {
   .movement-card{
-    display: flex;
-    flex-direction: column;
-    width: 100% !important;
-    margin: 0 !important;
+    width: 100%;
+    margin: 0;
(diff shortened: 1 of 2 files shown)


Drvanon:
What about a min width for screens that are bigger?

marickmanrho:
Nice suggestion. Will add. Any suggestion on what would be a good min width? Also at which screen size do we want full screen?

Drvanon:
I think you could also maybe make use of these helper classes: https://ionicframework.com/docs/layout/css-utilities#ionic-breakpoints

gridt/src/app/home/home.page.scss
```diff
@@ -5,4 +5,14 @@
...
     justify-content: space-between;
-}
\ No newline at end of file
+}
+
+// The width of cards on the home page should match the width of the screen for small enough screens. Issue #60
+@media screen and (max-width: 600px) {
+  ion-card{
+    display: flex;
+    flex-direction: column;
```
Drvanon, line 14: These characteristics are set by ionic already, why would we change them?

gridt/src/app/home/home.page.scss
```diff
@@ -5,4 +5,14 @@
...
+    width: 100% !important;
+    margin: 0 !important;
```
Drvanon, line 16: Why are you using !important here? 

(Generally speaking this is seen as bad practice if it is not *absolutely* necessary.)

gridt/src/app/home/home.page.scss
```diff
@@ -7,9 +7,13 @@
...
 }
 
+.movement-card{
+  min-width: 300px;
+}
+
 // The width of cards on the home page should match the width of the screen for small enough screens. Issue #60
-@media screen and (max-width: 600px) {
-  ion-card{
+@media screen and (max-width: 576px) {
```
Drvanon, line 15: I think it would be nice to use css variables here: https://ionicframework.com/docs/theming/css-variables#using-css, it leaves more flexibility and makes the code more understandable in my opinion.

gridt/src/app/home/home.page.scss
```diff
@@ -7,9 +7,13 @@
...
+  .movement-card{
     display: flex;
     flex-direction: column;
     width: 100% !important;
```
Drvanon, line 19: I did some testing on my device and it looks to me like the `!important`s are unnecessary now.

gridt/src/app/home/home.page.scss
```diff
@@ -7,9 +7,13 @@
...
+  min-width: 300px;
+}
+
 // The width of cards on the home page should match the width of the screen for small enough screens. Issue #60
-@media screen and (max-width: 600px) {
-  ion-card{
+@media screen and (max-width: 576px) {
+  .movement-card{
     display: flex;
     flex-direction: column;
```
Drvanon: I don't see/understand(?) the difference these two statements make, care to explain?

gridt/src/app/home/home.page.scss
marickmanrho: It would indeed be perfect. But, unfortunately css variables are not yet supported for '@media' queries

PR #67, opened 2020-04-18
Limit message to 140 characters on signaling.

marickmanrho:
When signaling on movement, limit message length to 140 characters.

### DIFF
--- a/gridt/src/app/home/home.page.ts
+++ b/gridt/src/app/home/home.page.ts
@@ -174,7 +174,7 @@ export class HomePage implements OnInit, OnDestroy {
     const el = await this.alertCtrl.create({
       header: "Want to send a message with your signal?",
       message: '',
-      cssClass: "confirmSignal_alert",
+      cssClass: "confirmSignal-alert", // makes message text go red
       inputs: [
         {
           name: "message",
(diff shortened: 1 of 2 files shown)


gridt/src/app/home/home.page.ts
```diff
@@ -190,7 +192,12 @@ export class HomePage implements OnInit, OnDestroy {
         {
           text: "Yes",
           handler: (data) => {
-            this.signal(movement, data.message);
+            if(data.message !==null && data.message.length < 140){
```
Drvanon, line 195: Could you set spaces around the if statements? It reads a little cramped. See line 162 for an example.

gridt/src/app/home/home.page.ts
```diff
@@ -190,7 +192,12 @@ export class HomePage implements OnInit, OnDestroy {
...
+              this.signal(movement, data.message);
+            }else{
```
Drvanon, line 197: Normally I leave spaces around the else.

gridt/src/app/home/home.page.ts
Drvanon, line 195: Right now, if the user has made given no message, but presses yes, it will simply send. Is this desired behaviour?

gridt/src/app/home/home.page.ts
```diff
@@ -173,6 +173,8 @@ export class HomePage implements OnInit, OnDestroy {
   async confirmSignal (movement: Movement) {
     const el = await this.alertCtrl.create({
       header: "Want to send a message with your signal?",
+      message: '',
+      cssClass: "confirmSignal_alert",
```
Drvanon, line 177: Css classes usually work with dashes (-) instead of lodashes (_).

gridt/src/app/home/home.page.ts
marickmanrho: Actually, !== means the message may not be empty. But if it is empty, we might want a custom error message.

PR #68, opened 2020-04-18
Disable menu when not logged in

Yori-O92:


gridt/src/app/app.component.ts
```diff
@@ -36,6 +40,10 @@ export class AppComponent implements OnInit, OnDestroy{
     private auth: AuthService
   ) {
     this.initializeApp();
+    this.isLoggedIn$=this.router.events.pipe(
```
Drvanon, line 43: Could you give the "=" some space to breath? Just put some spaces around it for readability's sake, please.

gridt/src/app/app.component.html
```diff
@@ -1,7 +1,7 @@
 <ion-app>
 
-    <ion-menu contentId="main" side="start" menuId="first">
-      <ion-header >
+    <ion-menu contentId="main" side="start" menuId="first" *ngIf="isLoggedIn$|async">
```
Drvanon, line 3: Same for the pipe character. Readability is crucial, especially in the HTML, where code often is just less pretty.

gridt/src/app/app.component.ts
Drvanon, line 43: It might be good to leave a comment explaining why it is necessary to do the piping, because that might not be completely clear from just looking at the code.

PR #70, opened 2020-04-18
Update README for guidelines on contribution

Drvanon:
Thought I would have it in writing as I have explained to quite a few members how this works now.


Drvanon:
Good call! I always mess that word up.

PR #74, opened 2020-04-19
Issue 50: create logout button

Yori-O92:


gridt/src/app/app.component.html
```diff
@@ -15,12 +15,13 @@
...
-        
-        <ion-button expand="full">
-          Logout
-        </ion-button>
+        </ion-list>      
       </ion-content>
+      <ion-footer>
+        <ion-button expand="full" (click)="auth.logout()" href="/">
+          Logout
+      </ion-button>
```
Drvanon, line 23: Nitpicking, but: this is not on the right indentation level.

gridt/src/app/app.component.html
```diff
@@ -15,12 +15,13 @@
...
           </ion-menu-toggle>
-        </ion-list>
-        
-        <ion-button expand="full">
-          Logout
-        </ion-button>
+        </ion-list>      
       </ion-content>
+      <ion-footer>
+        <ion-button expand="full" (click)="auth.logout()" href="/">
```
Drvanon, line 21: It might be interesting to make use of a nice icon here as well. I would recommend `<ion-icon name="exit-outline"></ion-icon>`.

Drvanon:
Would you mind merging into api-developement instead of master?

gridt/src/app/app.component.html
Drvanon: Also, for not wanting to attract too much attention to something we wish user use the least, maybe consider using color="light", to make it less attention drawing.

Yori-O92:
I changed the button color to 'dark', which apparently turns the text dark and the background color light (for contrast), which I think is most subtle.

PR #76, opened 2020-04-22
Issue 73

Yori-O92:
Changed user.model.ts to add the avatar attribute (string) and added the avatars to the homepage. Reduced size of leader names as well.

### DIFF
--- a/gridt/src/app/core/user.model.ts
+++ b/gridt/src/app/core/user.model.ts
@@ -16,5 +16,5 @@ export interface User {
     /**
      * The URL linking to the user's profile picture on gravatar.com
      */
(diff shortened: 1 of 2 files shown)


gridt/src/app/core/user.model.ts
```diff
@@ -13,5 +13,8 @@ export interface User {
         message?: string
     } | null;
 
+    /**
+     * The URL linking to the user's profile picture on gravatar.com
+     */
     avatar: string;
```
Drvanon, line 19: Currently the server does not respond with avatars yet, so maybe add a question mark, as to indicate that this isn't there yet?

PR #78, opened 2020-04-26
Add a privacy policy

### LINKED ISSUE #26: As a user I want to read a privacy policy such that I can asses if my privacy is governed
The privacy policy should be accessible without login and explicitly visible when signing up.

wiki description of 'privacy policy':
A privacy policy is a statement or a legal document (in privacy law) that discloses some or all of the ways a party gathers, uses, discloses, and manages a customer or client's data. It fulfills a legal requirement to protect a customer or client's privacy. Personal information can be anything that can be used to identify an individual, not limited to the person's name, address, date of birth, marital status, contact information, ID issue, and expiry date, financial records, credit information, medical history, where one travels, and intentions to acquire goods and services. In the case of a business it is often a statement that declares a party's policy on how it collects, stores, and releases personal information it collects. It informs the client what specific information is collected, and whether it is kept confidential, shared with partners, or sold to other firms or enterprises. Privacy policies typically represent a broader, more generalized treatment, as opposed to data use statements, which tend to be more detailed and specific.

marickmanrho:
We need a privacy policy to make sure we comply with European GDPR law. This commit adds a first version of a privacy policy which covers the basic needs to the main folder. 

Implementation into the app is still required. For example by displaying a cookie consent massage with a link to the privacy policy and through the about page. 

Fixes issue #26.


Drvanon:
We are not working with cookies, so we don't have to present them with that. It would indeed be good to have one of these, but we were thinking of talking to actual lawyers, because our needs are very specific.

Drvanon:
Also, how are you planning on showing this information to the user? A text in the source code is probably not enough, or?

Yori-O92:
I very much like the initiative on this. I just edited the file (quite rigorously). I agree with Robin that at some point, it may be wise to get some professional advice. However, I do believe most of this lies within our own hands. As information is our core business, we should be as aware as possible of what we collect, how we use it and how we not use data obtained.
About the cookies: our homepage does place cookies, which is part of the Wordpress service. So the privacy policy now reflects that. I was not sure about the 'log files', what are they and do they apply to what we do as well. I did leave the section in for now. I also added a few statements on what we do and don't do with data collected and adapted the part about third parties and partners to reflect our current situation and envisioned future.

Yori-O92:
In my opinion, the privacy policy should be made available on our homepage and should be approved by the user upon registering an account in the app.

marickmanrho:
> 
> 
> We are not working with cookies, so we don't have to present them with that. It would indeed be good to have one of these, but we were thinking of talking to actual lawyers, because our needs are very specific.

Alright, I thought at least the session key is saved locally.

@Yori-O92 Great changes, this makes it much better.  

I thought about linking to the privacy policy while making an account an on the settings page of the app. Also on the cookie message on the website and in the website footer.

Yori-O92:
I have now made it so that the privacy policy modal can be called on the register page. Also checking a box that states the approval of the privacy policy is now required before registering. I believe this makes the privacy part complete (save for future updates of the policy) and mergeable with api-development.

Yori-O92 (approved):
I think it's ok now.

gridt/src/app/login/register/register.page.html
```diff
@@ -67,6 +67,19 @@
                 Should at least be 6 characters long.
               </ion-label>
             </ion-item>
+            
+            <ion-item>
+              <ion-grid>
+                <ion-row>
+                  <input type="checkbox" [checked]="ch1" (change)="ch1= !ch1" style="padding: 2em">
```
Drvanon, line 74: Calling a variable "ch1" is not good practice, it is not clear what is meant by this.

gridt/src/app/login/register/register.page.ts
```diff
@@ -66,4 +68,12 @@ export class RegisterPage implements OnInit {
       .then(alertEl => alertEl.present());
   }
 
+  async presentPrivacy() {
+    const modal = await this.modalController.create({
+      component: PrivacyPage
+    });
+    console.log("present privacy policy");
```
Drvanon, line 75: In general we try to keep the standard log level as empty as possible to prevent cluttering. These type of messages normally get level "debug".

I would recommend a change to:
`console.debug("Present privacy policy");`

Drvanon (changes requested):

gridt/src/app/about/privacy/privacy.page.ts
```diff
@@ -0,0 +1,22 @@
...
+
+@Component({
+  selector: 'app-privacy',
+  templateUrl: './privacy.page.html',
+  styleUrls: ['./privacy.page.scss'],
+})
+export class PrivacyPage implements OnInit {
+
+  constructor(public modalController: ModalController) { 
+
```
Drvanon, line 12: Nit picking: In these cases I normally do `constructor () {}` instead of the redundant line breaks.

Yori-O92:
I implemented the requested changes.

Drvanon (commented):
Looks good, won't merge yet because the issue-18-new branch has not been merged yet, so I will merge when that is done.

I also have not tested yet, so that will have to happen still.

HendrikPreuter:
What is the status of this?

marickmanrho:
> 
> 
> What is the status of this?

Well, this is still waiting for the about page to be finished.

Yori-O92:
I will pick up this issue again once the about page is approved.

Issue #18, opened 2019-02-27
As a user I want to receive information on how the Gridt works, so that I can use the network more effectively.


PJvanVliet:

Drvanon:
I think this would be best implemented with a "slide show" of cards at the first opening off the app.

Yori-O92:
I pushed a branch to the repo with the current state of the code relating to this issue. There is now an additional page that explains what the app does (more design needed, though.) However, I'm not sure if the tutorial guard is functioning properly. This should display the 'about' page upon first time logging in, and should remember when the user has seen the tutorial so that it doesn't pop up again unasked. I'm using ionic storage (https://ionicframework.com/docs/angular/storage) to store the information, as instructed by this tutorial: https://fireship.io/lessons/ionic4-intro-slides-tutorial-to-educate-app-users/).
Maybe someone can give it a review? I will dig further to check if I can verify that it works correctly.

Yori-O92:
Fixing this issue on branch new-Issue-18 still requires the writing of tests and the addition of buttons that allow the user to move the slides forward and backward.

PR #79, opened 2020-04-26
Create tutorial/welcome page

Drvanon:
This would resolve issue #18.


gridt/src/app/about/about.page.scss
```diff
@@ -0,0 +1,25 @@
...
+
+ion-img {
+    max-width: 50vw;
+    max-height: 50vw;
+    overflow: hidden;
+}
+
+p {
+    padding: 1em;
+}
```
Drvanon, line 15: It is a good idea to limit addressing tags directly, instead use something like `ion-slides.intro-slides`, or something along those lines to make sure you don't change classes in other parts of the application, as well as keeping your intentions very clear.

gridt/src/app/about/tutorial.guard.spec.ts
```diff
@@ -0,0 +1,15 @@
...
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      providers: [TutorialGuard]
+    });
+  });
+
+  it('should ...', inject([TutorialGuard], (guard: TutorialGuard) => {
+    expect(guard).toBeTruthy();
+  }));
+});
```
Drvanon, line 15: There are issues in all of your tests. Please try to resolve these.

Yori-O92:
There is a test that fails, which has to do with the 'Tutorial page' routing. However, I don't really understand the error message or how to resolve this test failure.

Drvanon:
The about page and the privacy policy are separate issues. It would be sensible to treat them in different pull requests.

Drvanon:
Also, props for the file layout, I would have suggested that, but you were ahead of me.

Yori-O92:
> 
> 
> The about page and the privacy policy are separate issues. It would be sensible to treat them in different pull requests.

A wow you are fast. Thanks for the feedback. Indeed, the privacy policy is more pertaining to Issue #26. There's also a pull request for that, since we just put the privacy policy as a document in the repository. I think I should stage-commit-push-pull-request more often for every issue separately in the future, but I'm not sure how to fix the mixup between issues now, since I've already committed and pushed everything onto the #18 branch...

Drvanon:
This pull request was made obsolete by #85.

Issue #62, opened 2020-04-18
As a client I want to be able to update my profile.


Drvanon:
In the api/core there should be the possibility to call a function to perform an API call that updates the bio of the user. The user and signal models will have to be updated/created for this. 

This is necessary for #8.

Issue #51, opened 2020-04-08
As a user I want to have a settings page where I have a single place to communicate all my preferences.


Drvanon:
This page should contain the possibility to reset/change the password of the user (#27), set the profile of the user (#3/#8), and enable/disable sending/receiving nudges/kudos from my followers (#15/#16).

marickmanrho:
This Issue is currently on hold until there is a server call which can handle the request. On the server side this is issue 44

marickmanrho:
We will use a localstorage to store the settings. How do we treat this storage when the user logs out? Should it delete/reset or should it persist? 

In my opinion it makes sense to create two different storage locations. One secure storage to store the identity of the user. This secure storage is deleted when the user logs out. Then a normal local storage which stores the "Device Settings". These settings will be specific for the device in use and will persist on the device when logging out. This is odd, because if somebody else where to log in on the same device, the settings from the previous user will apply.

Actually, I think the best option would be to store all settings on the server. All settings will then be stored in a secure storage locally and on the server. Then we can just delete the local storage upon user logout and reload from the server upon login. The settings are now truly user-settings and do not carry over from previous users.

Drvanon:
Why should we not simply reset to default settings when a user logs out? User wide settings would have the effect that two devices can not have different settings. I think it would be good to discuss this with the team.

On the user identity settings: currently it is not possible to truly securely store information on the user's side, because we are using a progressive web app. There is an alternative way available to remove local storage entirely, but that would require rewriting authentication on both the server and client side entirely, which is maybe a little out of the scope of this particular issue, although it is on a project list somewhere here on Github.

Issue #27, opened 2019-03-23
As a user, I want to be able to change my password, in case I forget it


marickmanrho:
A password recovery option should be available. There should be three links available to reset your password. One on the user's profile page, one directly visible on the login page, and one when a login attempt fails three times in a row. 

The recover password page should ask for some form of authentication (like the user's email). If this email is in our system, the user should receive an email with a link. By clicking on this link the user is redirected to a page on which a new password can be chosen.

Drvanon:
Why is this issue in user profiles, @Yori-O92 ?

Yori-O92:
I considered it to be something that should be accessed via the individual
user settings, so I placed it under the 'profile' project. Although I would
also agree if it was placed somewhere else, as it relates to authentication.

Drvanon:
Maybe we could make a new project purely for authentication. I think I have a few issues that I would like to add to that as well.

Yori-O92:
OK!

Yori-O92:
I made a new project called 'authentication' on the client side and moved
the issue there. I did not create any new projects on the server side,
though.

Issue #16, opened 2019-02-26
As a user, I want to be able to disable and enable the sending and receiving of nudges, so that I am not bothered by followers I cannot see when I don't want to and so that I won't do the same to others.


Yori-O92:

marickmanrho:
This will be implemented on the settings page (see issue #51 )

PR #81, opened 2020-04-28
Add a settings page

marickmanrho:
This will add a settings page (#51 ) where the user can:
- Profile
    - Review user profile #3 
    - Edit user bio #8 
- Account settings
    - Review username and email
    - Change email and password #27 
- Gridt settings
    - Toggle receiving kudos #15 
    - Toggle receiving nudges #16 
    - Toggle receiving messages #14 

Also related is the API related Issue #62 .


gridt/src/app/core/lastSignal.model.ts
```diff
@@ -0,0 +1,8 @@
+export interface LastSiganl {
+
```
Drvanon, line 2: Spelling mistake

Drvanon: Could you call this just "signal"? It will be used in other contexts than just the last signal.

gridt/src/app/core/Signal.model.ts
```diff
@@ -0,0 +1,8 @@
+export interface Signal {
+
+//Last signal
+
```
Drvanon, line 4: Why is there so much whitespace?

AndraUdrescu: To not look clustered. Tho I do have the tendency to overuse white space

Drvanon: I would say that it is a little too much here.

gridt/src/app/settings/settings.page.html
```diff
@@ -24,11 +24,9 @@
     <ion-row justify-content-center>
       <ion-col size="12" size-sm="8" offset-sm="2" size-md="6" offset-md="3" text-center>
         <ion-card>
-
-          <!-- <img src="https://www.gravatar.com/avatar/HASH"> -->
           <ion-card-header>
-            <img src="https://www.gravatar.com/avatar/HASH">
-            <ion-card-title>John Doe</ion-card-title>
+            <img [src]="(identity$ | async)?.gravatar">
```
Drvanon, line 28: The gravatar will not be accessible in this manner. (This was a recent update on the server side.) Instead call md5 on identity.email. (With the appropriate async stuff of course.)

Drvanon: You might need to install a library for this or copy the code from css-tricks.

gridt/src/app/settings/settings.page.ts
```diff
@@ -1,15 +1,43 @@
...
 
+  async showError(error:string) {
+    const el = await this.alertCtrl.create({
+      header: "Something went wrong while obtaining your profile from the server.",
+      message: error,
+      buttons: ["Okay"]
+    });
+
+    el.present();
+  }
```
Drvanon, line 42: Code looks good and is very sensible, but it does not get called anywhere, as far as I could see. 

Thinking about the UX: how would a user try to perform a reload if it actually throws an error?

gridt/src/app/core/settings.service.ts
```diff
@@ -1,107 +1,136 @@
...
+import { SecureStorageService } from './secure-storage.service';
+
+@Injectable({
+  providedIn: "root"
+})
+export class SettingsService {
+  /*
+   * Subscribe to this observable to ready the API.
+   */
+  public URL = "https://api.gridt.org";
```
Drvanon, line 17: I think it would be logical to have this in the settings as well. But if we would make settings server specific this setting would be paradoxical.

gridt/src/app/core/settings.service.ts
```diff
@@ -1,107 +1,136 @@
...
+  private getServerSettings = this.auth.readyAuthentication$.pipe(
+      flatMap((options) => this.http.get<Identity>(
+        `${this.URL}/identity`,
+        options
+      )),
+      catchError( this.handleBadAuth(this.disabler$) ),
+      map( id => {
+        return {identity: id}
+      })
+    )
```
Drvanon, line 104: In my opinion this would be more suitable in api.service.ts and then imported here, but maybe api.service.ts should be changed instead.

marickmanrho: Sure, how to treat the disabler observable? Also move that to the api.service?

gridt/src/app/core/settings.service.ts
marickmanrho: This becomes redundant in settings.service if we move the getServerSettings function to the api.service. In the settings service a function should be added which gets the preferred server URL from local storage. This function can then be called from the api.service.

gridt/src/app/core/settings.service.ts
Drvanon: I think that that should go to settings.page.ts, because it is specifically about which UI elements and that should be handled by the page itself.

gridt/src/app/core/settings.service.ts
Drvanon: Getting the setting "which server should I address" from a server sounds a little strange to me, but I agree that once local storage is functional, this should be fixed.

gridt/src/app/core/settings.service.ts
marickmanrho: I plan to check against the disabler also when trying to store settings locally. Especially when other pages will also be able to change settings, the disabler function should stay inside the settings.service.

PJvanVliet:
The API has changed a little since starting, in order to clean up the endpoints a little.

- /change_password -> /user/change_password
- /change_email -> /user/change_email
- /request_password_reset -> /user/reset_password/request
- /reset_password -> /user/reset_password/confirm

PR #84, opened 2020-05-12
Make movement descriptions display linebreaks

Yori-O92:


gridt/src/app/movements/movements-detail/movements-detail.page.html
```diff
@@ -13,9 +13,9 @@
       <ion-row>
         <ion-col size-md="6" offset-md="3">
           <h2>{{ (movement$ | async)?.name }}</h2>
-          <p><strong>{{ (movement$ | async)?.short_description }}</strong></p>
-          <p *ngIf="(movement$ | async )?.description">{{ (movement$ | async)?.description }}</p>
-
+          <p class="shortdescription">{{ (movement$ | async)?.short_description }}</p>
```
Drvanon, line 16: Why did you remove the strong?

Yori-O92: The font type is now included in the css class for the shortdescription as well. It comes out the same.

Drvanon: That is a code design decision. I don't know how much you know about this, but if you are unfamiliar with semantic html, I would recommend reading the following article. I don't necessarily argue one way or the other, but I do think that it is useful to make it in an informed manner.

https://medium.com/adalab/the-importance-of-semantic-html-78e74fb75ff0

Yori-O92: Interesting consideration and good to become aware of. However, I think that in this case, the <strong> tag has a layout function than rather than a semantic one. Therefore, I would still leave it in the scss file.

Drvanon: Okay, then I'll approve your pull request.

PR #85, opened 2020-05-16
Create about page with tutorial and credits

### LINKED ISSUE #82: As a user I want to have an About page where I can find info about the Gridt project.
The About page should be a separate page available in the main menu.

The page should display information about the Gridt project:
- Gridt mission statement
- Basic contact information
    - Link to the [Gridt website](https://www.gridt.org)
- Contribute links to [OpenCollective](https://opencollective.com/gridt) and [Github](https://github.com/GridtNetwork/)
- A link to redo the tutorial (see also #18 ) such that a user can review information about how the Gridt app works.

Yori-O92:
This is the cleaned up version of the Issue-18 branch, with exclusively the code pertaining to issue 18


Yori-O92:
This pull request makes #79 obsolete.

marickmanrho:
Don't forget to change the branch to development!

Drvanon:
Do you want us to start reviewing yet?

Yori-O92:
No, this is old stuff. I still have to work on the navigation buttons and
the tests for the page. Will need some time still.

Yori-O92:
I finished the 'about page', with some tests included. On my computer, swiping and lay-out look good in different screens. It currently doesn't have navigation buttons in the slide show yet, but if we would like that in the future, I prefer fixing it in a new issue, as well as unnoticed layout flaws. (Everything can always be prettier.)

gridt/src/app/about/about.page.ts
```diff
@@ -0,0 +1,37 @@
...
+@Component({
+  selector: 'app-about',
+  templateUrl: './about.page.html',
+  styleUrls: ['./about.page.scss'],
+})
+export class AboutPage {
+
+  constructor(public modalController: ModalController) { 
+
+  }
```
Drvanon: Style thing: can you make this one line?

gridt/src/app/about/credits/credits.page.scss
```diff
@@ -0,0 +1,3 @@
+.padded {
+    padding: 1em;
+}
```
Drvanon: Isn't there an ion-class for this?

gridt/src/app/about/tutorial/tutorial.page.html
```diff
@@ -0,0 +1,68 @@
+<ion-content>
+  <ion-grid>
+    <!-- <ion-row *ngIf="router.url === '/about'" class="ion-float-right">
+      <ion-col  class="ion-float-right">
+        <ion-button (click)="finish()">
+          <ion-icon name="close" slot="icon-only"></ion-icon>
+        </ion-button>
+      </ion-col>
+    </ion-row> -->
```
Drvanon: Please do not commit commented out code. It needs to be clear if it is necessary yes or no, not half half.

gridt/src/app/about/tutorial/tutorial.page.ts
```diff
@@ -0,0 +1,28 @@
...
+  selector: 'app-tutorial',
+  templateUrl: './tutorial.page.html',
+  styleUrls: ['./tutorial.page.scss'],
+})
+export class TutorialPage implements OnInit {
+
+  constructor(private router: Router,
+              public modalController: ModalController) { }
+
+
```
Drvanon: That is one space too many.

gridt/src/app/about/tutorial/tutorial.page.ts
```diff
@@ -0,0 +1,28 @@
...
+  finish() {
+    if(this.router.url === "/about"){
```
Drvanon: The braces should have a bit more breathing room.

gridt/src/app/about/tutorial/tutorial.page.ts
```diff
@@ -0,0 +1,28 @@
...
+      this.modalController.dismiss()
+    }
+    else {
+      this.router.navigateByUrl('/movements')
```
Drvanon: You are forgetting some ';'s

gridt/src/app/home/home.page.html
```diff
@@ -7,26 +7,9 @@
...
-          </ion-button>
-        </p>
-      </ion-col>
-    </ion-row>
-  </ion-grid>
+<ion-content>
 
-  <!-- Can be rewritten as an if/else statement together with previous if-->
-  <ion-grid *ngIf="movements$ && (movements$ | async).length > 0">
+  <ion-grid *ngIf="movements$ && (movements$ | async).length > 0; else Tutorial">
```
Drvanon, line 12: This means that if the user is subscribed to only 1 movement and then unsubscribes he will see the tutorial again. Is this desired behaviour? (Not a rethorical question.)

gridt/src/app/home/home.page.html
```diff
@@ -76,8 +59,12 @@ <h3>
...
           </ion-card-content>
         </ion-card>
-    </ion-col>
-  </ion-row>
+      </ion-col>
+    </ion-row>
   </ion-grid>
 
+  <ng-template #Tutorial>
+    <app-tutorial></app-tutorial>
```
Drvanon, line 67: I don't think that the 'app' in `app-tutorial` contributes much to the understanding of what this component does. I would suggest keeping it to simply tutorial.

Drvanon:
Gave some code quality comments and general stuff. Still need to look at the effects. It looks a heck of a lot better than what it did before.

gridt/src/app/about/about.page.scss
```diff
@@ -0,0 +1,10 @@
+.padded {
```
marickmanrho: There is also the ionic `padding` option. [(see this link)](https://ionicframework.com/docs/layout/css-utilities) Did you consider using that? I can imagine you want a more specific padding, but in that case its better to rename `.padded` into something a bit more descriptive. For example `.small_padding`

gridt/src/app/about/about.page.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  // If the tutorial is opened, should open modal
+  it('should open the tutorial modal', () => {
+    component.presentTutorial();
+    expect(modalSpy.create).toHaveBeenCalled();
+  });
+  
+  // If roll credits is opened, should open modal
+  it('should open the credits modal', () => {
+    component.presentCredits();
```
marickmanrho, line 38: It may add to test the actual button on the page.

gridt/src/app/about/about.page.html
```diff
@@ -0,0 +1,23 @@
...
+      <ion-menu-button></ion-menu-button>
+    </ion-buttons>
+    <ion-title>About</ion-title>
+  </ion-toolbar>
+</ion-header>
+
+<ion-content>
+  <ion-grid>
+    <ion-row>
+      <ion-col size-sm="6" offset-sm="3">
```
marickmanrho, line 13: Maybe you could use `.ion-justify-content-center` to prevent setting the margins to auto manually in the css.

gridt/src/app/about/credits/credits.page.html
```diff
@@ -0,0 +1,23 @@
+<ion-header>
+  <ion-toolbar>
+    <ion-title>Credits</ion-title>
+  </ion-toolbar>
+</ion-header>
+
+<ion-content>
+  <div class="padded">
+    <p>This app was designed and prorammed by:</p>
```
marickmanrho: programmed

gridt/src/app/about/credits/credits.page.scss
marickmanrho: Yup, I've included a link to CSS Utilities in an other comment.

gridt/src/app/about/tutorial/tutorial.page.ts
marickmanrho: Not sure if I like that we're using an exception based return url. I was thinking of passing a return url to the modal form when you open it. Or you could split the router.url based on `/` and subtract the last term. It depends a bit if this is worth the change, is the tutorial going to be accessible only through the about page?

gridt/src/app/about/tutorial/tutorial.page.html
```diff
@@ -0,0 +1,68 @@
...
+
+    <ion-row>
+      <ion-col>
+        <ion-buttons *ngIf="router.url === '/about'">
```
marickmanrho, line 5: Why do you only want this button when entering from the about page? Why not always have the option to cancel the tutorial?

gridt/src/app/about/tutorial/tutorial.page.scss
```diff
@@ -0,0 +1,45 @@
...
+    overflow: hidden;
+}
+
+.intro-img {
+    max-width: 90vw;
+    max-height: 90vw;
+    overflow: hidden;
+}
+
+p {
```
marickmanrho: See comment about `ion.padding`

gridt/src/app/about/tutorial/tutorial.page.scss
```diff
@@ -0,0 +1,45 @@
...
+    text-align: justify;
+}
+
+.step-four {
+    height: 100vh;
+    flex-direction: column;
+    text-align: justify;
+}
+
+.step-five {
```
marickmanrho: This class seems to be the same for each step. You could make a single class named `.step` and add that to each step in the html.

gridt/src/app/about/tutorial/tutorial.page.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  }));
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
+
+  //It should navigate to movements if finish is clicked on homepage
+  it('should navigate to movements if finish is called', () => {
```
marickmanrho: Could you extend this test to also see if it goes to `/about` when needed.

marickmanrho (changes requested):
Great commit! Nice clean code and good implementation. Though there are some minor suggestions. I would like some navigation buttons on the tutorial. Now it is very clumsy to drag your mouse across the page on a large screen. I know you've already tried implementing them, so I would also be open to the suggestion to create a new Issue for that.

Yori-O92:
Thanks for the comments, will work on them tomorrow!

gridt/src/app/home/home.page.html
Yori-O92: Yes, it is intended this way, so that the home page is never empty. If the user decides to unsubscribe from all their movements, the tutorial just comes back, which will redirect to the movements page if finished.

gridt/src/app/home/home.page.html
Yori-O92: The app-'something' is the name of the component selector, so that this page appears here. For all pages, the selector starts with app, e.g. app-add-movement, or app-home. So I stuck to this convention also for the tutorial page.

gridt/src/app/about/tutorial/tutorial.page.html
Yori-O92: The idea is that if the tutorial appears on the home page, this is because the user is not subscribed to any movements, so the tutorial fills the home page. The user can always navigate away from the homepage if they want, but the idea is that the user finishes the tutorial and is then navigated to the movements page. If the user opens the tutorial from the about page, this can also happen by accident (selected the wrong button), or because they just want to get a quick look. So for those cases, I put the close button in the corner for easy dismissal.

gridt/src/app/about/tutorial/tutorial.page.scss
Yori-O92: Sharp. They were different colors before, but that didn't look good with the animations. So now it is indeed just one style.

gridt/src/app/about/tutorial/tutorial.page.html
marickmanrho: Then you probably also want to overwrite the router return address when coming from the home page. Because now if your home page is empty, and thus the tutorial is shown, what happens when you try and return to the previous page? This is now again the home page, and thus shows you the same tutorial page again. That may give the user the idea that he/she is stuck.

Yori-O92:
I will check it thoroughly tomorrow. The way it is now is that if the
homepage is empty, the tutorial component gets loaded in the homepage. So
the router is still set at 'home', except that the content that is
displayed is from the tutorial. So if the user then navigates back, they
will land at the page where they were before 'home'. But I'll make sure to
check it.

gridt/src/app/about/tutorial/tutorial.page.html
Yori-O92: I will check it thoroughly tomorrow. The way it is now is that if the
homepage is empty, the tutorial component gets loaded in the homepage. So
the router is still set at 'home', except that the content that is
displayed is from the tutorial. So if the user then navigates back, they
will land at the page where they were before 'home'. But I'll make sure to
check it.

Yori-O92:
There's a test that is failing, in which I'm trying to mock the router state, depending on which the 'finishTutorial' function should do different things. However, mocking the router state is a lot harder than I thought. If anyone has any ideas for this piece of code, let me know.

Drvanon:
Pretty bad gif, but I have this on the tutorial page:
output
Basically, the page is larger than the screen creating a scroll bar effect. I don't know why this is the case on my computer, but it seems like something you did not intend for to happen.

gridt/src/app/about/tutorial/tutorial.page.html
Drvanon: You haven't gotten back to him yet.

Drvanon (commented):
I reallly like the looks of what you have made. The pager is very sweet as well. The point I have is the height issue that I mentioned.

PR #93, opened 2020-06-10
Create CI.yml

Classper:
Only echoes output out this point, not calling testscripts yet


Drvanon:
It's pretty sweet that I also immediately can see where it failed. Interestingly I can not even merge it right now. I wonder if that is because of the check or because it is only a draft...

Drvanon:
Are we going to fix the tests in this PR or shall we move that to a seperate one?

Issue #100, opened 2020-08-19
As a developer I want no tests to fail, so that I can do TTD effectively.


Drvanon:

marickmanrho:
I think this is actually fixed already, @Drvanon do you agree?

PR #101, opened 2020-08-20
Fix tests

Drvanon:
This pull request is intended to fix all standing issues with tests and should resolve them.

#100

### DIFF
--- a/gridt/src/app/app.component.html
+++ b/gridt/src/app/app.component.html
@@ -1,28 +1,28 @@
 <ion-app>
-    <ion-menu contentId="main" side="start" menuId="first" *ngIf="isLoggedIn$ | async">
-      <ion-header>
-        <ion-toolbar>
-          <ion-title>Menu</ion-title>
-        </ion-toolbar>
-      </ion-header>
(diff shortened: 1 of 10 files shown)


gridt/src/app/app.component.html
```diff
@@ -1,28 +1,28 @@
...
+      </ion-list>
+    </ion-content>
+    <ion-footer>
+      <ion-button color="dark" expand="full" (click)="auth.logout()" href="/">
+        <ion-icon class="ion-padding-end" name="exit-outline"></ion-icon>
+        Logout
+      </ion-button>
+    </ion-footer>
+  </ion-menu>
+  <ion-router-outlet id="main"></ion-router-outlet>
```
marickmanrho, line 27: Why is this file in the commit? There seem to be no changes.

marickmanrho (approved):
Perfect!

gridt/src/app/app.component.html
Drvanon: I believe the indentation was incorrect which I resolved here.

Issue #107, opened 2020-09-01
As a user, I want to be able to confirm or cancel the activation of my signal after I press the 'activate button'


Yori-O92:

Yori-O92:
This was already solved before.

PR #108, opened 2020-09-01
Add cancel option to signal dialog instead of 'no signal message'

Yori-O92:
This closes issue #107


Yori-O92:
Wait, I already made a PR for this once, so this was completely unnecessary.

PR #112, opened 2020-09-03
Issue 51 Settings Page

marickmanrho:
This will create a settings page and setting service. It also adds the required API functionality in the API service to change the user's bio, email and password.

This will add a settings page (#51 ) where the user can:
Profile
 - Review user profile #3 
 - Edit user bio #8 

Account settings
 - Review username and email
 - Change email and password #27 

Gridt settings
 - Toggle receiving kudos #15 
 - Toggle receiving nudges #16 
 - Toggle receiving messages #14

Also related is the API related Issue #62 .


gridt/src/app/core/settings.service.ts
```diff
@@ -0,0 +1,137 @@
...
+        return {...settings.server}
+        // Server settings have priority over local settings
+      })
+    ).subscribe( (set) => this.secStore.set$("settings", {set}).subscribe() );
+  }
+
+  /**
+   * Update the _user_settings$ by combining Local and Server responses.
+   */
+  public getUserSettings(): void {
```
Drvanon, line 119: Why does this function exist if `get the_user_settings$ ()` also does? Most likely this is not the optimal name for this variable.

gridt/src/app/settings/change-email/change-email.page.html
```diff
@@ -0,0 +1,59 @@
+<ion-content class="ion-padding">
```
Drvanon, line 1: I have not been able to review this work, as I could not load the server response on my device.

gridt/src/app/settings/change-email/change-email.page.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+      providers: [
+        { provide: AuthService, useClass: AuthServiceStub }
+      ]
+    }).compileComponents();
+
+    fixture = TestBed.createComponent(ChangeEmailPage);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  }));
+
```
Drvanon, line 66: I could imagine that testing that there is a sensible map between the settings that are stored and how they are displayed could help making sure that it works.

gridt/src/app/settings/change-email/change-email.page.ts
```diff
@@ -0,0 +1,75 @@
...
+  ngOnInit() {
+  }
+
+  public closePopover() {
+    this.popoverCntrl.dismiss({
+      'dismissed': true
+    });
+  }
+
+
```
Drvanon, line 38: Why the white space?

gridt/src/app/settings/change-email/change-email.page.ts
```diff
@@ -0,0 +1,75 @@
...
+  async write_email(form: NgForm) {
+    this.edit_email$ = true;
+
+    const el = await this.loadingCtrl.create({
+      message: 'Updating email address...'
+    });
+
+    el.present();
+
+    this.api.postEmail$(form.value.password, form.value.email).pipe(timeout(500)).subscribe(
```
Drvanon, line 51: Why do you wait half a second?

gridt/src/app/settings/settings.page.html
```diff
@@ -0,0 +1,177 @@
...
+    </ion-title>
+  </ion-toolbar>
+</ion-header>
+
+
+<ion-content padding>
+  <!-- Refreshes page if pulled down (mobile only)-->
+  <ion-refresher slot="fixed" (ionRefresh)="refreshPage($event)" class="ion-hide-md-up">
+    <ion-refresher-content></ion-refresher-content>
+  </ion-refresher>
```
Drvanon, line 17: Cool! UX wise: how can we make users aware of this feature?

gridt/src/app/settings/settings.page.html
```diff
@@ -0,0 +1,177 @@
...
+        <ion-card>
+          <ion-card-header>
+            <ion-card-title>Profile</ion-card-title>
+          </ion-card-header>
+
+          <ion-card-content>
+            <ion-list>
+              <ion-item class="ion-text-center" lines="none">
+                <ion-avatar class="ion-text-center">
+                  <img [src]="gravatar">
```
Drvanon, line 31: When the server is unavailable this has a weird effect of creating a call to `/null`. Consider wrapping this in an if statement.

Drvanon (changes requested):
Part of the intention of building a settings page was to allow the setting of the API URL through the application itself instead of in the code. I have not seen this applied in the API requests in core/api.service.ts, which seems to still be calling `this.URL`.

gridt/src/app/core/settings.service.spec.ts
```diff
@@ -0,0 +1,253 @@
...
+  it('should fail to read local settings when not logged in', () => {
+    expect(secStoreStub.get$).not.toHaveBeenCalledWith('settings');
+  })
+
+  it('should fail to update local settings when not logged in', () => {
+    // Make sure there are some settings available
+    service._user_settings$.next(mock_settings[0])
+
+    // Create subscription to localstorage
+    service.the_user_settings$;
```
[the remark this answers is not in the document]
marickmanrho: It actually does something. It make sure there is a subscription of the _user_settings which writes changes into the localstorage.

gridt/src/app/core/user.model.ts
```diff
@@ -7,7 +9,7 @@ export interface User {
      * The user's username
      */
     username: string;
-
+  
     last_signal: {
         time_stamp: string;
         message?: string
```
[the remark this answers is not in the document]
marickmanrho: Good point, this has been quite a while ago...

gridt/src/app/core/user.model.ts
```diff
@@ -1,3 +1,5 @@
+import { Signal } from './Signal.model';
+
 export interface User {
```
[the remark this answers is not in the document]
marickmanrho: Signals can be personal too. It can also be my own signals in a movement.

gridt/src/app/core/settings.service.ts
```diff
@@ -0,0 +1,137 @@
...
+    private http: HttpClient,
+    private auth: AuthService,
+    private secStore: SecureStorageService,
+    private api: ApiService
+  ) {}
+
+  /**
+   * Create ReplaySubject which yields the latest user settings.
+   */
+  public _user_settings$ = new ReplaySubject<Settings>(1);
```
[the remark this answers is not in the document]
marickmanrho: I did this for initial testing purposes. Will change in next update. Create a get and set method instead and keep _user_settings$ private.

gridt/src/app/core/settings.service.ts
marickmanrho: Will change to update_user_settings, that's better.

gridt/src/app/core/Signal.model.ts
```diff
@@ -1,5 +1,4 @@
 export interface Signal {
-//Signal
-    time_stamp: string;
-     message?: string;
-}
\ No newline at end of file
+      time_stamp: string;
+      message?: string;
```
Drvanon, line 3: The indentation level here is still 4.

gridt/src/app/core/api.service.ts
```diff
@@ -276,15 +276,15 @@ export class ApiService {
       );
     }
 
-    public postPassword$( old_password: string, new_password: string ) {
+    public changePassword$( old_password: string, new_password: string ) {
       console.debug(`Saving new password to the server. (at leat it should now create a http.post)`);
```
Drvanon, line 280: I am not sure if the whole at least phrase is useful to begin with honestly.

gridt/src/app/core/settings.service.spec.ts
```diff
@@ -137,35 +137,30 @@ describe("IdentityService_AuthSuccesfull", () => {
   });
 
   // Local storage
-  it('should create localStorage upon first retreival of settings', () => {
+  it('should create localStorage upon first retrieval of settings', () => {
     // Subscribe local storage
-    service.the_user_settings$.subscribe();
+    service.the_user_settings$;
```
Drvanon, line 142: This should theoretically do nothing, right?

gridt/src/app/settings/change-password/change-password.page.ts
```diff
@@ -47,7 +47,7 @@ export class ChangePasswordPage {
 
     el.present();
 
-    this.api.postPassword$(form.value.old_password, form.value.new_password).pipe(timeout(500)).subscribe(
+    this.api.changePassword$(form.value.old_password, form.value.new_password).pipe(timeout(500)).subscribe(
```
Drvanon: Here a timeout of half a second remains, is there a reason for that?

Drvanon (commented):
Quick review of the last changes.

gridt/src/app/core/settings.service.spec.ts
marickmanrho: Good point, will fix in next commit.

gridt/src/app/settings/change-password/change-password.page.ts
marickmanrho: Nope, will remove in next commit.

gridt/src/app/core/api.service.ts
```diff
@@ -228,4 +232,59 @@ export class ApiService {
...
+    * Observable to obtain identity from server
+    */
+   public getServerIdentity$ = this.auth.readyAuthentication$.pipe(
+    flatMap((options) => this.http.get<Identity>(
+      `${this.URL}/identity`,
+      options
+    )),
+    catchError( this.handleBadAuth() ),
+    map( id => {
+      return {identity: id} // Return a <Settings> compatible object
```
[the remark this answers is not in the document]
marickmanrho: I don't really see the benefit of this yet. Remember that the Settings will later be expanded to include notification settings also.

gridt/src/app/core/settings.service.spec.ts
```diff
@@ -0,0 +1,253 @@
...
+  });
+
+});
+
+describe("IdentityService_AuthFailed", () => {
+
+  class authServiceStub_failed {
+    isLoggedIn$ = of(false);
+    readyAuthentication$ = throwError("Failed Authentication");
+  }
```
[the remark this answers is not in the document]
marickmanrho: To provide a better readability I would suggest breaking the rules here.

gridt/src/app/core/settings.service.spec.ts
```diff
@@ -0,0 +1,253 @@
...
+    service.the_user_settings$;
+    expect(secStoreStub.set$).not.toHaveBeenCalledWith('settings')
+  });
+
+  it('should remove localStorage upon bad auth.', () => {
+    // Simulate login
+
+    // Simulate logout
+
+    // Make sure localstorage is empty
```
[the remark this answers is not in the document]
marickmanrho: Actually, this test belongs to the SecStore. Therefore I suggest making a Issue and write this test in an other pull request.

gridt/src/app/core/settings.service.ts
```diff
@@ -0,0 +1,137 @@
...
+}
+
+@Injectable({
+  providedIn: "root"
+})
+export class SettingsService {
+  /*
+   * Subscribe to this observable to ready the API.
+   */
+  public URL = "https://api.gridt.org";
```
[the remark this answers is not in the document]
marickmanrho: Fixed the first comment. The second comment I think should be part of a new Issue. This is not functionality which we try to implement now.

gridt/src/app/settings/settings.page.html
marickmanrho: I was thinking about maybe a special notification for mobile devices. But that I want to do in a future Issue.

gridt/src/app/core/settings.service.spec.ts
```diff
@@ -0,0 +1,253 @@
...
+  afterEach(() => {
+    httpMock.verify();
+  });
+
+  // Basic functionality
+  it('should be created', () => {
+    expect(service).toBeTruthy();
+  });
+
+  it('should provide the last available settings', () => {
```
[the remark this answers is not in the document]
marickmanrho: It is not completely redundant as it is a good base test to make sure the whole train is working as it is supposed to be.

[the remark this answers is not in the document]
marickmanrho: But I see your point about actually testing rxjs at the moment. I'll try to make it a marble testing diagram and implement `jasmine_marbles.hot()`

gridt/src/app/settings/change-email/change-email.page.spec.ts
marickmanrho: Am I right in thinking this comment relates more to the settings page instead of the `change-email.page` since this page is only used as a modal.

gridt/src/app/core/settings.service.spec.ts
```diff
@@ -0,0 +1,253 @@
...
+  });
+
+  // !!!!!!!!!!!!
+  // Not completely clear why this is needed??? (It's also in api.service.spec)
+  // !!!!!!!!!!!!
+  afterEach(() => {
+    httpMock.verify();
+  });
+
+  it('should fail to read local settings when not logged in', () => {
```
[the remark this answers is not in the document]
marickmanrho: This is tested to a lower degree in `api.service.spec` by `it(should fail to read settings from server when not logged in)`

gridt/src/app/core/settings.service.spec.ts
Drvanon: This does something: when I remove it, the test fails. However you are not subscribing you are running the 'get' function of this variable. Maybe it would be good to check here that this is equal to some expected value, especially when it has not been set before.

gridt/src/app/core/settings.service.spec.ts
```diff
@@ -180,14 +170,34 @@ describe("IdentityService_AuthSuccesfull", () => {
...
+    // Obtain the settings
+    service.updateUserSettings();
+
+    // Make sure the local storage settings are returned
+    expect(service.the_user_settings$).toBeObservable(cold('a',{a: mock_settings[0]}))
+  })
+
 });
 
 describe("IdentityService_AuthFailed", () => {
```
Drvanon, line 203: Sometimes my test suite throws an error when loading this test suite and some times it does not:
pic-selected-200907-1007-50

gridt/src/app/core/settings.service.ts
```diff
@@ -17,44 +17,46 @@ export interface ServerMessage {
   providedIn: "root"
 })
 export class SettingsService {
-  /*
-   * Subscribe to this observable to ready the API.
-   */
-  public URL = "https://api.gridt.org";
 
   constructor (
     private http: HttpClient,
```
Drvanon, line 22: This dependency has become redundant because you have properly stored all api calls in the api service. This means you can remove reference of this here, as well as in your imports and in the tests, cleaning it up nicely.

gridt/src/app/core/settings.service.ts
```diff
@@ -17,44 +17,46 @@ export interface ServerMessage {
...
+  };
 
   /**
    * Transforms the user settings Subject to an Observable
    */
   get the_user_settings$ (): Observable<Settings> {
     // If the user_settings change, make sure the new settings are stored
     // in the localstorage.
+    let cachedValue: Settings;
     if (this.auth.isLoggedIn$.subscribe()) {
```
Drvanon, line 56: This if statement will always pass, because the `Observable.subscribe` always returns a subscription. I could suggest piping the `isLoggedIn$` observable. Also right now you are not making use of the cachedValue variable.

gridt/src/app/core/settings.service.ts
```diff
@@ -17,44 +17,46 @@ export interface ServerMessage {
...
-      // filter(val => !!val ),  // Skip no changes in the server setting
-    );
+  public set_user_settings(settings): void {
+    this._user_settings$.next(settings);
+  };
 
   /**
    * Transforms the user settings Subject to an Observable
    */
   get the_user_settings$ (): Observable<Settings> {
```
Drvanon, line 52: Right now this function is not returning anything, so this will always be empty. This should be something that is tested.

gridt/src/app/core/settings.service.ts
Drvanon: By the fact that all tests pass, I can tell that this is not a feature that is tested at the moment. I would recommend writing the test first and then implementing this feature.

gridt/src/app/core/settings.service.ts
```diff
@@ -68,17 +70,22 @@ export class SettingsService {
...
-    return of(false)
-    // return this.disabler$.asObservable();
+    return this.disabler$.asObservable();
+  }
+
+  private Dissable(disabler) {
+    return function(error) {
+      disabler.next(true)
+      return throwError(error);
+    };
```
Drvanon, line 80: Why is this an observable? I think in this particular case a lot of simplification could be done by simply letting this be a variable, but maybe I do not understand in sufficient depth why you chose to do this. If there is a good reason, a comment would be in order.

gridt/src/app/core/settings.service.ts
```diff
@@ -68,17 +70,22 @@ export class SettingsService {
...
   }
 
   /**
    * Observable to obtain settings from secure storage
+   * If secure storage is empty, return empty settings object
    */
-  private getLocalSettings$ = this.auth.readyAuthentication$.pipe(
-   flatMap(() => this.secStore.get$("settings")),
-   catchError( this.handleBadAuth(this.disabler$) )
+  private getLocalSettings$: Observable<Settings> = this.auth.readyAuthentication$.pipe(
```
Drvanon, line 87: Naming convention: this is not a function, so `localSettings$` sounds to me like a more intuitive name for this.

gridt/src/app/core/settings.service.ts
```diff
@@ -87,120 +94,35 @@ export class SettingsService {
    */
   private storeLocalSettings(settings: Settings): void {
     console.log(`Storing ${JSON.stringify(settings)} into the local storage.`);
```
Drvanon, line 96: Printing out an object in the logs is not necessary. Maybe also debug is a more suitable level for this statement as well as some other logs you are making.

Drvanon: I would actually recommend doing a search for 'log' and revisiting the necessity of all of them.

gridt/src/app/core/settings.service.ts
```diff
@@ -87,120 +94,35 @@ export class SettingsService {
...
-      })
-    ).subscribe( (set) => this.secStore.set$("settings", {set}).subscribe() );
+    this.secStore.set$("settings", settings).subscribe();
   }
 
   /**
    * Update the _user_settings$ by combining Local and Server responses.
+   *
+   * While a forkJoin may be usefull, it doesn't allow for simple error
+   * catching, hence the two seperate subscriptions for local and server.
```
Drvanon, line 104: I am not yet entirely convinced, why could you not catch the errors before they enter the forkJoin?

Drvanon: You have done so successfully for the secureStorage which seems to work quite well.

gridt/src/app/settings/change-bio/change-bio.module.ts
```diff
@@ -0,0 +1,20 @@
+import { NgModule } from '@angular/core';
```
Drvanon, line 1: Question: why is there an entire folder and module for a single component? Can this not be resolved by simply adding the component in the settings folder?

gridt/src/app/settings/change-bio/change-bio.page.ts
```diff
@@ -0,0 +1,73 @@
...
+        el.dismiss();
+        this.closePopover();
+      },
+      (error) => {
+        el.dismiss();
+        this.showError(error);
+      }
+    );
+
+    timer(500).subscribe( () => this.SetService.updateUserSettings());
```
Drvanon, line 61: Why the .5 second wait?

Drvanon: If you want to make sure that the server settings are the same as those stored locally, subscribing to the request makes more sense. If that doesn't error, you can be pretty sure that it has arrived properly, if you still want to check it would make sense to do the request at that point, but .5 seconds is pretty arbitrary, especially on phones that do not always have the best connections.

gridt/src/app/settings/settings.page.html
Drvanon: Is it not the case that the settings always return _something_? In that case this would do nothing. Writing a test for this would not be a bad idea.

gridt/src/app/settings/change-password/change-password.page.ts
```diff
@@ -40,12 +42,12 @@ export class ChangePasswordPage {
...
 
     const el = await this.loadingCtrl.create({
-      message: 'Updating email address...'
+      message: 'Updating password...'
     });
 
     el.present();
 
-    this.SetService.postPassword$(form.value.old_password, form.value.new_password).pipe(timeout(500)).subscribe(
+    this.api.changePassword$(form.value.old_password, form.value.new_password).pipe(timeout(2500)).subscribe(
```
Drvanon, line 50: What is the reason for the 2.5s wait?

gridt/src/app/settings/settings.page.ts
```diff
@@ -29,19 +32,23 @@ export class SettingsPage implements OnInit  {
...
     console.log(`gravatar is ${JSON.stringify(this.gravatar)}`)
-    if (this.isDisabled$) {
-      this.serverWarning()
-    }
+
+    // Raise warning toast when isDisabled$ becomes true
+    this.isDisabled$.pipe(
+      filter (val => val == true),
+      tap( () =>  this.serverWarning() )
+    ).subscribe();
```
Drvanon, line 51: This can be simplified just a bit:

```ts
this.isDisabled$.pipe(
  filter( val => val )
).subscribe( () => this.serverWarning() );
```

gridt/src/app/settings/settings.page.ts
```diff
@@ -50,17 +57,9 @@ export class SettingsPage implements OnInit  {
...
-    event.target.complete();
-  }
-
-  public refreshPage() {
-    this.SetService.getUserSettings();
-    if (this.isDisabled$) {
-      this.serverWarning()
-    }
+  public refreshPage(event?) {
+    this.SetService.updateUserSettings();
```
Drvanon, line 61: Fun feature that could be built here: if you know you are waiting for a server call, you can connect a loading spinner to the observable of the request. That helps the user understand that he is waiting "for the internet". Not necessary and maybe hard to implement right now, but it could be nice.

Drvanon: Actually the fact that it is hard to implement tells me something is off. Probably here you actually do not wish to call `updateUserSettings`, as it also pulls from the localStorage, what you really want is to just make the server call here.

gridt/src/app/settings/change-email/change-email.page.ts
```diff
@@ -57,7 +58,7 @@ export class ChangeEmailPage{
       }
     );
 
-    timer(500).subscribe( () => this.SetService.getUserSettings());
+    timer(500).subscribe( () => this.SetService.updateUserSettings());
```
Drvanon, line 61: In a different comment I state why I don't think that this timer is a good idea.

gridt/src/app/core/settings.service.ts
marickmanrho: The idea is to use disabler$ in the html of the settings page to disable certain UI elements. Since the disabler$ may change value while loading the page, I thought it to be wise making it an observable.

marickmanrho: But maybe you're right, I now see how a variable could be just fine.

marickmanrho:
This pull request is no longer needed. The PR has been cut up into different smaller ones.

Drvanon:
Woop!

PR #125, opened 2020-09-07
Issue 113 center cards

HendrikPreuter:
Changes

Added:
- row / col switched
- offset and size added for medium+ size screens
- style changed for some padding around each card

Removed:
- style width elements as the card should not take up the full screen, but looks nicer with a bit of padding, even on smaller devices.

Problem:
- seems to have some issue with iphone5 still.

### DIFF
--- a/gridt/package-lock.json
+++ b/gridt/package-lock.json
@@ -2837,7 +2837,6 @@
       "version": "6.10.2",
       "resolved": "https://registry.npmjs.org/ajv/-/ajv-6.10.2.tgz",
       "integrity": "sha512-TXtUUEYHuaTEbLZWIKUr5pmBuhDLy+8KYtPYdcV8qC+pOZL+NKqYwvWSRrVXHn+ZmRRAu8vJTAznH7Oag6RVRw==",
-      "dev": true,
       "requires": {
         "fast-deep-equal": "^2.0.1",
         "fast-json-stable-stringify": "^2.0.0",
@@ -2887,8 +2886,7 @@
     "ansi-regex": {
       "version": "2.1.1",
       "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-2.1.1.tgz",
-      "integrity": "sha1-w7M6te42DYbg5ijwRorn7yfWVN8=",
-      "dev": true
+      "integrity": "sha1-w7M6te42DYbg5ijwRorn7yfWVN8="
     },
     "ansi-styles": {
       "version": "3.2.1",
@@ -3015,7 +3013,6 @@
       "version": "0.2.4",
       "resolved": "https://registry.npmjs.org/asn1/-/asn1-0.2.4.tgz",
       "integrity": "sha512-jxwzQpLQjSmWXgwaCZE9Nz+glAG01yF1QnWgbhGwHI5A6FRIEY6IVqtHhIepHqI7/kyEyQEagBC5mBEFlIYvdg==",
-      "dev": true,
       "requires": {
         "safer-buffer": "~2.1.0"
       }
@@ -3061,8 +3058,7 @@
     "assert-plus": {
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/assert-plus/-/assert-plus-1.0.0.tgz",
-      "integrity": "sha1-8S4PPF13sLHN2RRpQuTpbB5N1SU=",
-      "dev": true
+      "integrity": "sha1-8S4PPF13sLHN2RRpQuTpbB5N1SU="
     },
     "assign-symbols": {
       "version": "1.0.0",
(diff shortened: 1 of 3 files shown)


gridt/src/app/home/home.page.scss
```diff
@@ -10,11 +10,8 @@
...
-
-// The width of cards on the home page should match the width of the screen for small enough screens. Issue #60
-@media screen and (max-width: 576px) {
-  .movement-card{
-    width: 100%;
-    margin: 0;
-  }
+.movement-card
+{
+     margin: 10px;
```
marickmanrho, line 15: The indentation is inconsistent in this document. The standard in this project is an identation of 2.

gridt/package-lock.json
```diff
@@ -1083,168 +1083,68 @@
       }
     },
     "@angular/pwa": {
-      "version": "0.901.1",
```
marickmanrho, line 1086: I recommend against updating the package-lock.json file in these small updates since that makes it hard to track when dependencies have updated.

gridt/src/app/home/home.page.html
```diff
@@ -27,8 +27,8 @@
 
   <!-- Can be rewritten as an if/else statement together with previous if-->
   <ion-grid *ngIf="movements$ && (movements$ | async).length > 0">
-    <ion-col>
-      <ion-row>
```
marickmanrho, line 31: By changing the order each card is now full column width. This is fine for mobile devices but does look a bit awkward on larger screens. Is there something you can do to improve the visuals of the cards on medium to large screens?

marickmanrho (changes requested):
Great first pull request! It sure fixes the Issue, but there are some points which need a little attention.

gridt/src/app/home/home.page.html
HendrikPreuter: If you mean the cards themselves then i think a new issue is in order as that falls outside of the scope of this issue i think.
Edit: The cards themselves are not of equal size when the amount of leaders differ.

HendrikPreuter: If you mean displaying multiple items in one row, than that can be realized by moving the ngFor to the col instead of the item. This fixes that somewhat while still working for smaller devices. The problem i am now encountering is that the offset is getting repeated.

Drvanon (commented):
I agree with all of Marick's comments and would like to encourage the use of the media-query classes, as you have done in home.page.html:31.

HendrikPreuter (commented):
I will check to see if the content is centered with just 1 item on the page. Since my latest test where all with four items

gridt/package-lock.json
marickmanrho: Yes, you can make a revert of a single file with `git checkout <commit_hash> -- <file>`

marickmanrho (approved):
Looks like a clean solution to the problem.

Drvanon:
Completely agree with Marick that the solution feels very clean, quite a fan.

I just did a little check, because I wanted to confirm the behaviour for different screen sizes. For big screens it behaves perfectly, but for the smallest sizes I don't think I am fully satisfied yet. I think it might be out of the scope of this issue (might just make a new issue), but if you have a quick _and beautiful_ solution we could fit it in here.
pic-selected-200907-1855-41

HendrikPreuter:
> Completely agree with Marick that the solution feels very clean, quite a fan.
> 
> I just did a little check, because I wanted to confirm the behaviour for different screen sizes. For big screens it behaves perfectly, but for the smallest sizes I don't think I am fully satisfied yet. I think it might be out of the scope of this issue (might just make a new issue), but if you have a quick _and beautiful_ solution we could fit it in here.
> pic-selected-200907-1855-41

Is that still the iphone 5? I Haven't been apble to get that specific one to work nicely just yet. Tho now that you mention that is the smallest screen i may have an idea how to fix this:
setting offset-xs="0" & size-xs="12"

HendrikPreuter:
> Completely agree with Marick that the solution feels very clean, quite a fan.
> 
> I just did a little check, because I wanted to confirm the behaviour for different screen sizes. For big screens it behaves perfectly, but for the smallest sizes I don't think I am fully satisfied yet. I think it might be out of the scope of this issue (might just make a new issue), but if you have a quick _and beautiful_ solution we could fit it in here.
> pic-selected-200907-1855-41

image
This is what i meant.

Drvanon:
Let me confirm this on my own machine, if it works, I will merge it immediately.

gridt/src/app/home/home.page.scss
```diff
@@ -8,13 +8,6 @@
...
-  min-width: 300px;
-}
-
-// The width of cards on the home page should match the width of the screen for small enough screens. Issue #60
-@media screen and (max-width: 576px) {
-  .movement-card{
-    width: 100%;
-    margin: 0;
-  }
+    margin: 5px;
```
Drvanon, line 11: Love the simplification!

Drvanon (approved):

Issue #106, opened 2020-08-25
As a User, I want to receive a warning about the safety of the app, so that I will choose a unique password.


marickmanrho:
When registering, changing password and resetting the password the user should receive a notification warning them that the security of the app is not good. Their password may be easy to steal.

PR #127, opened 2020-09-07
Issue #106 Add safety warning in register

HendrikPreuter:
Added:
-Warning added when entering the register page #106
-Alert tests on entering and on leaving
-Added dismiss in ngOnDestroy()

This is a short term change that should be removed when we feel more confident in our security and have implemented a better terms of service


marickmanrho:
It may be good to write a test, which makes sure the alert is indeed presented.

HendrikPreuter:
> It may be good to write a test, which makes sure the alert is indeed presented.

That is a good idea, how should i go about this? This way of testing is still what unfamiliar to me.

HendrikPreuter:
> > It may be good to write a test, which makes sure the alert is indeed presented.
> 
> That is a good idea, how should i go about this? This way of testing is still what unfamiliar to me.

The tests have been implemented.

Issue #64, opened 2020-04-18
As a follower I want to be able to see the profile and recent messages of my leaders from the home page so that I can make an informed decision if I want to swap this leader.


Drvanon:
When a user taps on a leader from the home page it should display a modal that shows the username, profile picture and biography of that leader, as well as a list of the recent messages this leader has sent.

PR #130, opened 2020-09-08
Issue #123 view full message 

### LINKED ISSUE #123: As a User, when my leaders type a long message i want to be able to read all of it.
When a leader types a longer message it gets cut off on smaller devices, therefore i want to be able to click on the message to extend it's text view in some manner to see the complete message.

HendrikPreuter:
Issue #123 

Done:
-expandable on click
-Tests

The message is still decoupled from the leader icon to allow for a modal to come in to show more details about the leader (ie. message history, bigger profile pic, maybe short bio) Issue #64


Drvanon:
This should be a modal, not an alert.

gridt/src/app/home/home.page.html
```diff
@@ -72,7 +72,7 @@ <h3>
                         </ion-text>
                           <app-expandable [expandedHeight]="itemExpandedHeight" [expanded]="itemExpanded">
                             <ion-text *ngIf="itemExpanded">
-                              <p class="overflow-visible" size="auto">{{ leader.last_signal.message }}</p>
+                              <p class="overflow-visible">{{ leader.last_signal.message }}</p>
```
Drvanon: I personally really liked the idea of having to tap on the message to see it. That would mean making this a conditional statement.

Drvanon:
Do you want feedback already?

gridt/src/app/home/home.page.html
HendrikPreuter: That should not be there as it does nothing.
And i agree with you, that is what it does already!

gridt/src/app/home/home.page.html
```diff
@@ -59,9 +59,23 @@ <h3>
...
+                        <!-- <ion-button fill="clear" (click)="showFullMessage(leader.last_signal.message)">
+                          <div class="leader-message">
+                            <small>
+                              {{ leader.last_signal.message }}
+                            </small>
+                          </div>
+                        </ion-button> -->
+                        <ion-text *ngIf="!itemExpanded">
+                          <p class="overflow-elipsis">{{ leader.last_signal.message }}</p>
+                        </ion-text>
```
HendrikPreuter: This piece shows the cut-off version of the text. if the text is clicked showFullMessage is executed which toggles itemExpanded which then in turn disables this code and enables the part you comented on below.

HendrikPreuter (commented):
> Do you want feedback already?

If you want to, feel free. Tho i am still working on the tests

gridt/src/app/home/expandable/expandable.component.html
```diff
@@ -0,0 +1,5 @@
+<div class='expand-wrapper' [class.expanded]="itemExpanded" (click)="toggleFullMessage()">
+  <ion-text>
+    <p class="overflow-visible">{{signalMessage}}</p>
```
Drvanon: Style comment: `{{ signalMessage }}` is nicer than `{{signalMessage}}`.

gridt/src/app/home/expandable/expandable.component.scss
```diff
@@ -0,0 +1,58 @@
...
+@media screen  and (min-width: 801px) and (max-width: 1500px) {
+        .expanded {
+                min-height: 120px;
+        }
+} 
+@media screen  and (min-width: 1501px) {
+        .expanded {
+                min-height: 70px;
+        }
+}
```
Drvanon: Generally speaking it is nicer to solve this with the classes from ionic, though if there is no other option this works, it's just hard to do maintenance on.

Drvanon: Hmm, it might not be possible to do it differently.

gridt/src/app/home/expandable/expandable.component.ts
```diff
@@ -0,0 +1,24 @@
+import { Component, OnInit, Input } from "@angular/core";
+
+@Component({
+  selector: "app-expandable",
+  templateUrl: "./expandable.component.html",
+  styleUrls: ["./expandable.component.scss"],
+})
+export class ExpandableComponent implements OnInit {
```
Drvanon: I think a more appropriate name for this could be chosen. The `Component` in a component class is generally used to indicate that it _is_ a component, not to describe what type of component it is. A more name that would fit convention better could be something like `SignalMessageComponent`.

gridt/src/app/home/expandable/expandable.component.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  ngOnInit() {}
+
+  async toggleFullMessage() {
+    if(this.signalMessage.length > 10) {
+      this.itemExpanded = !this.itemExpanded;
+    }
+  }
+
+
```
Drvanon: Redundant white space.

gridt/src/app/home/expandable/expandable.component.ts
```diff
@@ -0,0 +1,24 @@
...
+export class ExpandableComponent implements OnInit {
+
+  @Input() signalMessage;
+  itemExpanded = false;
+
+  constructor() { }
+
+  ngOnInit() {}
+
+  async toggleFullMessage() {
```
Drvanon: Why is this async?

gridt/src/app/home/expandable/expandable.component.ts
```diff
@@ -0,0 +1,24 @@
...
+  styleUrls: ["./expandable.component.scss"],
+})
+export class ExpandableComponent implements OnInit {
+
+  @Input() signalMessage;
+  itemExpanded = false;
+
+  constructor() { }
+
+  ngOnInit() {}
```
Drvanon: You can leave these two functions as well as the implementation of ngOnInit entirely.

gridt/src/app/home/expandable/expandable.component.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+    fixture = TestBed.createComponent(ExpandableComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  }));
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
+
+  it('should toggle itemExpanded', () => {
```
Drvanon: Good test.

gridt/src/app/home/expandable/expandable.component.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+    spyOn(component, 'toggleFullMessage');
+
+    let button = fixture.debugElement.query(By.css('.expand-wrapper')).nativeElement;
+    button.click();
+
+    fixture.whenStable().then(() => {
+      expect(component.toggleFullMessage).toHaveBeenCalled();
+    });
+    done();
+  });
```
Drvanon: Excellent! You are starting to understand the tests!

gridt/src/app/home/expandable/expandable.component.spec.ts
Drvanon: If you are testing this, it would be sensible to test that it also does _not_ do this when the text is too short.

gridt/src/app/home/expandable/expandable.component.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  async toggleFullMessage() {
+    if(this.signalMessage.length > 10) {
```
Drvanon: This is rather arbitrary choice. Are you able to find out if the message is being wrapped or not and then use that information to perform the check?

gridt/src/app/home/home.page.html
```diff
@@ -59,9 +59,24 @@ <h3>
...
+                          </div>
+                        </ion-button> -->
+                        <!-- <ion-text *ngIf="!itemExpanded">
+                          <p class="overflow-elipsis">{{ leader.last_signal.message }}</p>
+                        </ion-text>
+                          <app-expandable [signalMessage]="leader.last_signal.message">
+                            <ion-text *ngIf="itemExpanded">
+                              <p class="overflow-visible">{{ leader.last_signal.message }}</p>
+                            </ion-text>
+                          </app-expandable> -->
```
Drvanon: General rule of thumb: never commit commented code. It serves no purpose as your history is already captured by git.

gridt/src/app/home/home.page.scss
```diff
@@ -1,3 +1,7 @@
+.signal {
+
+}
```
Drvanon: This does not achieve anything.

gridt/src/app/home/expandable/expandable.component.scss
```diff
@@ -0,0 +1,58 @@
...
+@media screen  and (min-width: 601px) and (max-width: 800px) {
+        .expanded {
+                min-height: 100px;
+        }
+} 
+
+@media screen  and (min-width: 801px) and (max-width: 1500px) {
+        .expanded {
+                min-height: 120px;
+        }
```
Drvanon: Is this even necessary? When I click on the message, I want to see the message. It should not take up more space than necessary.

gridt/src/app/home/expandable/expandable.component.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  it('should show the expanded view', () => {
+    component.signalMessage = "this message is long enough to require more space.";
+    component.toggleFullMessage();
+    expect(component.itemExpanded).toBe(true, 'true after toggle');
+    let expandedEl = fixture.debugElement.query(By.css('.expand-wrapper, .expanded')).nativeElement;
+    expect(expandedEl.innerHTML).not.toBeNull();
```
Drvanon: Good idea to test this.

gridt/src/app/home/home.page.spec.ts
```diff
@@ -1,6 +1,6 @@
 import { of } from 'rxjs';
```
Drvanon: Changes here are unecessary.

gridt/src/app/home/home.page.ts
```diff
@@ -14,6 +14,7 @@ import { SwapService } from '../core/swap.service';
 })
 export class HomePage implements OnInit, OnDestroy {
   movements$ = new Observable<Movement[]>();
+  itemExpanded = false;
```
Drvanon: This should be removed.

Drvanon (changes requested):
You are definitely reaching the final stages of this commit. I like the way it is designed very much now, but there is some cleanup to do.

gridt/src/app/home/expandable/expandable.component.ts
HendrikPreuter: Shouldn't be, leftover from an earlier version. Good catch.

gridt/src/app/home/expandable/expandable.component.scss
HendrikPreuter: I agree that it shouldn't take up more room than necessary. That's why i added so many different versions. So that on each screen type the wasted space on expansion is minimized.

gridt/src/app/home/expandable/expandable.component.scss
HendrikPreuter: I agree, however i haven't found a way just yet. But i can fix the other issues and leave this over the weekend and just keep this pull request unresolved.

gridt/src/app/home/expandable/expandable.component.scss
Drvanon: Isn't the overflow visible class what makes the ... appear? Maybe you could just remove that class conditionally.

HendrikPreuter:
```
.overflow-ellipsis {
        text-overflow: ellipsis;
        white-space: normal;
        overflow: hidden;
    }
```
Should provide the '...' however they don't appear for me, whatever i do.
```
.overflow-visible {
        white-space: initial;
        
    }
```

Is what makes it that the text isn't cut off and continue on a new line.
If i remove the part commented on in the scss file than yes, it still expands. However the smooth transitional animation is thereby also removed as that only seems to work if the min-height changes.

So then the question becomes do we want easier to maintain or a nicer view experience for the user?

gridt/src/app/home/expandable/expandable.component.spec.ts
HendrikPreuter: When i change this test:

```
  it('should toggleFullMessage on click', (done) => {
    spyOn(component, 'toggleFullMessage');
    component.signalMessage = "this message is long enough to require more space.";
    component.width = 5;

    let button = fixture.debugElement.query(By.css('.expand-wrapper')).nativeElement;
    button.click();

    fixture.whenStable().then(() => {
      expect(component.toggleFullMessage).toHaveBeenCalled();
      expect(component.itemExpanded).toBe(true, 'true after toggle');
    });
    done();
  });
```

it fails on this:

`expect(component.itemExpanded).toBe(true, 'true after toggle');`

But i don't know why

HendrikPreuter: The new toggle function:
```
  ngOnInit() {
    this.width = document.getElementById('signalText').offsetWidth;
  }

  toggleFullMessage() {
    if (this.signalMessage.length > this.width) {
      this.itemExpanded = !this.itemExpanded;
    }
  }
```

gridt/src/app/home/expandable/expandable.component.scss
Drvanon: Partial solution: 
```scss
.expand-wrapper p {
        transition:max-height 0.5s;
        max-height: 16px;
        height: auto;
        overflow: hidden;
        text-overflow: ellipsis;
}
.expanded p {
        max-height: 300px;
        white-space: normal;
}
```

gridt/src/app/home/expandable/expandable.component.spec.ts
Drvanon: https://hatoum.com/blog/2016/11/12/jasmine-unit-testing-dont-forget-to-callthrough

marickmanrho (approved):
I think the solution presented is very elegant.

gridt/src/app/home/home.page.html
```diff
@@ -59,9 +59,9 @@ <h3>
                         </h3>
                         <p>{{ isLeaderDone(leader, movement) ? 'Done!' : 'Pending...' }}</p>
                       </div>
-                      <small *ngIf="isLeaderDone(leader, movement) && leader.last_signal.message">
-                        {{ leader.last_signal.message }}
-                      </small>
+                      <div *ngIf="isLeaderDone(leader, movement) && leader.last_signal.message" id="expandable">
+                        <app-expandable signalMessage={{leader.last_signal.message}}></app-expandable>
```
Drvanon: I would like there to be a different name here

gridt/src/app/home/expandable/expandable.component.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+    component.width = 5;
+
+    let button = fixture.debugElement.query(By.css('div')).nativeElement;
+    button.click();
+
+    fixture.whenStable().then(() => {
+      expect(component.toggleFullMessage).toHaveBeenCalled();
+      expect(component.itemExpanded).toBe(true, 'true after toggle');
+    });
+    done();
```
Drvanon: This done should be in the promise, now it isn't doing anything.

gridt/src/app/home/expandable/expandable.component.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+    component.signalMessage = "this message is short enough to not require more space.";
+
+    let button = fixture.debugElement.query(By.css('div')).nativeElement;
+    button.click();
+
+    fixture.whenStable().then(() => {
+      expect(component.toggleFullMessage).toHaveBeenCalled();
+      expect(component.itemExpanded).toBe(false, 'not toggled');
+    });
+    done();
```
Drvanon: Same for this done

Drvanon (commented):
Really good commit, it's starting to come together beautifully. I have to still look at the functionality when I get out of the hospital, but the code is looking very solid!

gridt/src/app/home/home.page.html
HendrikPreuter: Is leadersMessage a better name?

HendrikPreuter: Resolved via what's app.

gridt/src/app/home/expandable/expandable.component.ts
```diff
@@ -0,0 +1,22 @@
+import { Component, OnInit, Input } from "@angular/core";
+
+@Component({
+  selector: "signal-message",
+  templateUrl: "./expandable.component.html",
+  styleUrls: ["./expandable.component.scss"],
+})
+export class SignalMessageComponent implements OnInit{
+
+  @Input() signalText;
```
Drvanon: Within the context of our application I think a more suitable name is message text. Very banal change, but I think it helps comprehension.

gridt/src/app/home/expandable/expandable.component.scss
```diff
@@ -0,0 +1,17 @@
+.p {
```
Drvanon: I need a computer to be able to check out the functionality of this. But i won't be home until Thursday. Maybe you could make me a gif so I can see it?

gridt/src/app/home/expandable/expandable.component.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+    fixture.detectChanges();
+  }));
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
+
+  it('should toggle itemExpanded', () => {
+    component.signalText = "this message is long enough to require more space.";
+    component.width = 5;
```
Drvanon: By setting the width yourself you are not checking the width determining code you have created. Is it possible to set the screen width in the test frame work?

gridt/src/app/home/expandable/expandable.component.ts
```diff
@@ -0,0 +1,22 @@
+import { Component, OnInit, Input } from "@angular/core";
```
Drvanon: The file and folder name should be renamed still.

gridt/src/app/home/home.page.html
```diff
@@ -59,9 +59,9 @@ <h3>
                         </h3>
                         <p>{{ isLeaderDone(leader, movement) ? 'Done!' : 'Pending...' }}</p>
                       </div>
-                      <small *ngIf="isLeaderDone(leader, movement) && leader.last_signal.message">
-                        {{ leader.last_signal.message }}
-                      </small>
+                      <div *ngIf="isLeaderDone(leader, movement) && leader.last_signal.message" id="expandable">
+                        <signal-message signalText={{leader.last_signal.message}}></signal-message>
```
Drvanon: This element has only one real input and it does not necessarily need to be named, so what you could strive for (unless you have a good counter argument) is:

`<signal-message>{{ leader.last_signal.message }}</signal-message>`

gridt/src/app/home/expandable/expandable.component.spec.ts
HendrikPreuter: That is true! I hadn't thought of that. I will be keeping these tests i think and try and write 2 more to specifically test that if it can be set.

gridt/src/app/home/expandable/expandable.component.ts
HendrikPreuter: Why is that? Every time someone talked about giving a message to the followers it was called signaling. So why is signal message worse then message text?

gridt/src/app/home/expandable/expandable.component.ts
HendrikPreuter: this has given me some issues yesterday, but i will try again today.

gridt/src/app/home/expandable/expandable.component.scss
HendrikPreuter: yes, i will send you one soon.

HendrikPreuter (commented):
I will be processing the feedback today.

gridt/src/app/home/home.page.html
HendrikPreuter: that is what i had initially, but then i needed to switch over all of  the functionality for the switching the view to the signal message component. If i were to do it like this i will have to replace the <p> in the signal message component with <ng-content> and i am not sure the styling would work with that as good as it does now.

gridt/src/app/home/expandable/expandable.component.ts
HendrikPreuter: Turns out, i forgot to redo ionic serve...

gridt/src/app/home/expandable/expandable.component.scss
HendrikPreuter: It is in the slack/client

gridt/src/app/home/expandable/expandable.component.spec.ts
HendrikPreuter: So far, i haven't found a way to do this.
However even if we get the full screen width, that doesn't get us there as we need the width of <p>

HendrikPreuter: I could expect width to be bigger than 0, that way we know the function in onInnit did something. But that doesn't help much as it could still be a wrong value.

gridt/src/app/home/home.page.html
Drvanon: I feel like something went wrong when you typed this message. I also don't think I completely understand the argument, would you care to elaborate?

PR #133, opened 2020-09-08
Issue #121 Solve bug by dismissing in onDestroy()

### LINKED ISSUE #121: Modals and dialogs do not close when navigating 'back'
Replication:
- Navigate to home page
- Hit 'yes' on a subscribed movement
- While the dialog is open, navigate back using the 'back' button of phone or browser
- The app navigates to the previous page, but the overlay dialog/modal stays open.

HendrikPreuter:
Issue #121

Done:
- Dismiss in onDestroy()
-Alert dismiss test
-Find other instances of this issue

### DIFF
--- a/gridt/src/app/home/home.page.spec.ts
+++ b/gridt/src/app/home/home.page.spec.ts
@@ -19,6 +19,8 @@ describe('HomePage', () => {
   let fixture: ComponentFixture<HomePage>;
   let apiSpy: ApiService;
   let alertSpy: AlertController;
+  let alertSpy2: AlertController = jasmine.createSpyObj("alertSpy", ["create", "dismiss"]);
+  
 
   beforeEach(() => {
     jasmine.clock().install();
@@ -48,7 +50,8 @@ describe('HomePage', () => {
       providers: [
         { provide: ApiService, useValue: apiSpy },
         { provide: AlertController, useValue: alertSpy },
-        { provide: AuthService, useClass: AuthServiceStub }
+        { provide: AuthService, useClass: AuthServiceStub },
+        { provide: AlertController, useValue: alertSpy2}
       ]
     }).compileComponents();
 
@@ -63,6 +66,11 @@ describe('HomePage', () => {
 
   it('should create', () => {
     expect(component).toBeTruthy();
+  });
+
+  it('should dismiss the alert when leaving the page', () => {
+    component.ngOnDestroy();
+    expect(alertSpy2.dismiss).toHaveBeenCalled();
   });
 
   it('should find proper last signals', () => {
--- a/gridt/src/app/home/home.page.spec.ts
+++ b/gridt/src/app/home/home.page.spec.ts
@@ -18,8 +18,7 @@ describe('HomePage', () => {
   let component: HomePage;
   let fixture: ComponentFixture<HomePage>;
   let apiSpy: ApiService;
-  let alertSpy: AlertController;
-  let alertSpy2: AlertController = jasmine.createSpyObj("alertSpy", ["create", "dismiss"]);
+  let alertSpy: AlertController = jasmine.createSpyObj("alertSpy", ["create", "dismiss"]);
   
 
   beforeEach(() => {
@@ -39,9 +38,6 @@ describe('HomePage', () => {
       }
     );
 
-    alertSpy = jasmine.createSpyObj('alertSpy', {
-      create: new Promise( resolve => resolve({present: () => {}}) )
-    });
 
     TestBed.configureTestingModule({
       declarations: [ HomePage ],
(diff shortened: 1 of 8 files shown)


gridt/src/app/home/home.page.ts
```diff
@@ -26,7 +26,7 @@ export class HomePage implements OnInit, OnDestroy {
     this.api.getSubscriptions();
   }
 
-  ngOnDestroy() { }
+  ngOnDestroy() {this.alertCtrl.dismiss(); }
```
Drvanon, line 29: It is neater to do not use oneliners, as future git diffs will have to delete the entire line and replace it, while not actually changing the code. It is less readable as well.

HendrikPreuter: so would this be better?
image
or am i misunderstanding you?

Yori-O92:
I believe this is indeed what Robin means. Note also that this problem needs fixing in other pages as well, including the dialog box that pops up when you subscribe to or unsibscribe from a movement. There might be other instances as well.

gridt/src/app/home/home.page.ts
Drvanon: That is how I would like it.

Drvanon:
> I believe this is indeed what Robin means. Note also that this problem needs fixing in other pages as well, including the dialog box that pops up when you subscribe to or unsibscribe from a movement. There might be other instances as well.

I think it then would be sensible to include those instances in this pull request.

Drvanon:
I think you are still missing the test for the add-movement page.

HendrikPreuter:
> I think you are still missing the test for the add-movement page.

It should be there.

gridt/src/app/home/home.page.spec.ts
```diff
@@ -19,6 +19,8 @@ describe('HomePage', () => {
   let fixture: ComponentFixture<HomePage>;
   let apiSpy: ApiService;
   let alertSpy: AlertController;
+  let alertSpy2: AlertController = jasmine.createSpyObj("alertSpy", ["create", "dismiss"]);
```
Drvanon, line 22: Having two does not make sense, as one of them is not being used.

gridt/src/app/movements/movements-detail/movements-detail.page.spec.ts
```diff
@@ -1,30 +1,38 @@
...
+import { async, ComponentFixture, TestBed } from "@angular/core/testing";
+import { RouterTestingModule } from "@angular/router/testing";
+import { HttpClientModule } from "@angular/common/http";
+<<<<<<< HEAD
+=======
 
-import { MovementsDetailPage } from './movements-detail.page';
-import { AuthService } from 'src/app/core/auth.service';
-import { of } from 'rxjs';
+>>>>>>> fe1feac8813acb9cbd3248540d796f9397e807e6
```
Drvanon, line 8: I don't think you meant to have this here. Did you have issues with git?

gridt/src/app/movements/movements-detail/movements-detail.page.spec.ts
```diff
@@ -1,30 +1,38 @@
-import { CUSTOM_ELEMENTS_SCHEMA } from '@angular/core';
-import { async, ComponentFixture, TestBed } from '@angular/core/testing';
-import { RouterTestingModule } from '@angular/router/testing';
-import { HttpClientModule } from '@angular/common/http';
+import { CUSTOM_ELEMENTS_SCHEMA } from "@angular/core";
+import { async, ComponentFixture, TestBed } from "@angular/core/testing";
+import { RouterTestingModule } from "@angular/router/testing";
+import { HttpClientModule } from "@angular/common/http";
+<<<<<<< HEAD
+=======
```
Drvanon, line 6: This would most likely not compile or pass the tests. Please run the tests before committing.

gridt/src/app/movements/movements.page.spec.ts
```diff
@@ -50,4 +52,14 @@ describe('MovementsPage', () => {
   it('should create', () => {
     expect(component).toBeTruthy();
   });
+
+<<<<<<< HEAD
+  it('should dismiss the alert when leaving the page', () => {
+=======
+
+  it("should dismiss the alert when leaving the page", () => {
+>>>>>>> fe1feac8813acb9cbd3248540d796f9397e807e6
```
Drvanon, line 61: You did not really deal with this conflict I am getting the impression

Drvanon (changes requested):

gridt/src/app/movements/movements-detail/movements-detail.page.spec.ts
HendrikPreuter: That is strange... I did run the tests before my last commit and all of them passed.

gridt/src/app/movements/movements-detail/movements-detail.page.spec.ts
HendrikPreuter: Yes, i did indeed have some trouble.

gridt/src/app/movements/movements.page.spec.ts
HendrikPreuter: Aperantly not, no...

Drvanon (approved):

PR #136, opened 2020-09-08
Fix tests so that all of them now work

HendrikPreuter:


Drvanon:
Why was this relevant?

PR #139, opened 2020-09-08
Update CI.yml

Drvanon:
The old branch was no longer in use so for this to be here does not make much sense.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -7,9 +7,9 @@ name: CI
 # api-development/master later)
 on:
   push:
-    branches: [ master, api-development ]
+    branches: [ master ]
   pull_request:
-    branches: [ master, api-development ]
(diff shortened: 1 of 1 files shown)


Yori-O92 (approved):

.github/workflows/CI.yml
```diff
@@ -7,9 +7,9 @@ name: CI
 # api-development/master later)
 on:
   push:
-    branches: [ CI ]
+    branches: [ master, api-development ]
```
marickmanrho, line 10: This should be changed to development,

marickmanrho (changes requested):

.github/workflows/CI.yml
Drvanon: Good call.

marickmanrho (approved):
Great, ready to merge.

PR #140, opened 2020-09-08
Create FUNDING.yml

marickmanrho:
This will create a funding button in GitHub which points to our OpenCollective page.


marickmanrho:
Pulling directly to master since this is an effect only visible on GitHub.

Issue #55, opened 2020-04-08
It might be appropriate to split the movement and user models up into types that inherit of a base and specify for each situation.


Drvanon:
Right now we use one model of movements for all circumstances. However when we create a movement, the movement does not have an id. In the home page, it is important to keep track of how many replacements have happened this interval, etc. This adds information to the models that is only necessary in some cases. 

It would be good to specify these cases and inherit from a base case.

Drvanon:
Maybe you could take a look at this idea as well while implementing #63?

marickmanrho:
Some work has been done in #81 regarding the issue of models.

PR #141, opened 2020-09-09
Improve models

marickmanrho:
Removes:

- models from core folder
- signal from movement and user model

Adds:
- seperate model folder
- create new signal model

Fixes:
- #55


Drvanon:
First: I really like the new naming convention. It suits the situation a lot better. Second: I really like the smaller PRs it is just so much nicer to look at. Third: I really want the CI on these PRs, I find myself looking through the code, wanting to write a comment and needing to check it out locally first to run the tests, it'd be so much nicer to have that just straight up here. Fourth: I am getting a little sick of all the white space correction in these commits. I think they are redundant. Instead, we should just have one big correction of all files concerning white space and then deploy the new changes on top of that.

Drvanon:
The tests have come back positive, so I think I will accept this branch, but will then ask you to do the white space correction immediately after this.

Drvanon:
Could you move ServerMessage from api.core.ts also to a separate model?

marickmanrho:
Sure thing, just wait with the merger then. I'll update right away.

Drvanon:
Question: we were thinking of splitting Identity and Leader up in different interfaces. Do you want to do that in this PR or in another one?

marickmanrho:
Another, The Identity and Leader are not yet relevant at this stage. Will add at later stage.

gridt/src/app/core/models/credentials.model.ts
```diff
@@ -0,0 +1,8 @@
+export interface Credentials {
+  username: string;
+  password: string;
+}
+
+export interface AccessToken {
+  access_token: string;
+}
```
Drvanon, line 8: I would recommend referring to this file with "server-responses", as it describes both

gridt/src/app/core/models/servermessage.model.ts
```diff
@@ -0,0 +1,3 @@
+export interface ServerMessage {
+  message: string;
+}
```
Drvanon, line 3: If you will make the "server-response" file, this should be included there.

Drvanon (changes requested):
I think I would like to see it organized in a specific way, but am open to your opinion.

gridt/src/app/core/models/credentials.model.ts
marickmanrho: I did change this, but left Credentials seperate as it is not a server-response.

marickmanrho:
Should be good now.

Drvanon (approved):
Great. Merging it.

PR #143, opened 2020-09-09
Implement Identity API

### LINKED ISSUE #62: As a client I want to be able to update my profile.

marickmanrho:
Adds:

- Identity model
- Identity$() function to retrieve identity from server
- changeBio() to update user bio.
- changePassword() to update user password.

Does not add:
- changeEmail() functionality to avoid current vulnerability on server side.

Fixes #62

### DIFF
--- a/gridt/src/app/core/api.service.spec.ts
+++ b/gridt/src/app/core/api.service.spec.ts
@@ -115,7 +115,7 @@ describe("ApiService_failed_auth", () => {
 
   it('should fail to read identity from server when not logged in', () => {
     expect(
-      service.userIdentity$()
+      service.userIdentity$
     ).toBeObservable(cold("#", {}, "Can't authenticate: no credentials"));
   });
 
@@ -300,7 +300,7 @@ describe("ApiService_succesful_auth", () => {
   it('should be able to retreive identity', () => {
     httpClientStub.get.and.returnValue(of(mock_identity[0]));
 
-    expect(service.userIdentity$()).toBeObservable(cold('(a|)',{a: mock_identity[0]}));
+    expect(service.userIdentity$).toBeObservable(cold('(a|)',{a: mock_identity[0]}));
   });
 
   it('should be able to update the user bio', () => {
--- a/gridt/src/app/core/api.service.spec.ts
+++ b/gridt/src/app/core/api.service.spec.ts
@@ -1,109 +1,155 @@
-import {Type } from "@angular/core";
+import { Type } from "@angular/core";
 import { TestBed } from "@angular/core/testing";
+import { HttpClient, HttpHeaders, HttpErrorResponse } from '@angular/common/http';
 import { HttpClientTestingModule, HttpTestingController } from "@angular/common/http/testing";
 
-import { forkJoin, of, throwError } from "rxjs";
-import { skip, take, toArray } from "rxjs/operators";
+import { cold } from 'jasmine-marbles';
+import { of, throwError } from "rxjs";
 
+import { AuthService } from './auth.service';
 import { ApiService } from "./api.service";
+
 import { Movement } from "./models/movement.model";
 import { Identity } from "./models/identity.model";
 import { User } from './models/user.model';
-import { AuthService } from './auth.service';
 
-import { cold } from 'jasmine-marbles';
-import { HttpClient, HttpHeaders, HttpResponse, HttpHeaderResponse, HttpErrorResponse } from '@angular/common/http';
+let mock_movements: Movement[] = [
+  {
+    id: 10129312983,
+    name: "Mock Name",
+    short_description: "Mocking the short description",
+    description: "I will mock description. You are a fool, description.",
+    subscribed: false,
+    interval: "daily"
+  },
+  {
+    id: 1283912983123,
+    name: "Another mocked movement",
+    short_description: "This one only has a short description and no long one",
+    subscribed: false,
+    interval: "twice daily"
+  },
+  {
+    id: 10023123,
+    name: "Subscribed movement",
+    short_description: "This movement is supposed to come from the server.",
+    description: "We need some way to fake a response from the server, telling the client that it \
+     is subscribed to this particular movement",
+    subscribed: true,
+    interval: "daily"
+  },
+  {
+    id: 112312983,
+    name: "Flossing",
+    short_description: "Floss every day",
+    description: "We floss every day because it is good for our theeth.",
+    subscribed: true,
+    interval: "daily"
+  }
+];
 
-let mock_movements: Movement[];
 let mock_subscriptions: Movement[];
-let mock_identity: Identity[];
 
-describe("ApiService", () => {
-  let service: ApiService;
-  let auth: AuthService;
-  let httpMock: HttpTestingController;
-  let httpClientStub: jasmine.SpyObj<HttpClient>;
-  let authServiceStub: jasmine.SpyObj<AuthService>;
+let mock_identity: Identity[] = [
+  {
+    id: 1,
+    username: "ALittleOne",
+    bio: "Less then one foot tall.",
+    email: "a_little@one.com",
+    avatar: "arandomstring"
+  }
+];
 
-  const default_headers = {
-    headers: new HttpHeaders({
-      Authorization: "JWT aksdajskd.asdjknaskdn.asdjknakdnasjd"
-    })
-  };
+const default_headers = {
+  headers: new HttpHeaders({
+    Authorization: "JWT aksdajskd.asdjknaskdn.asdjknakdnasjd"
+  })
+};
 
+let service: ApiService;
+let auth: AuthService;
+let httpMock: HttpTestingController;
+let httpClientStub: jasmine.SpyObj<HttpClient>;
+let authServiceStub: jasmine.SpyObj<AuthService>;
+
+class authServiceStub_succes {
+  isLoggedIn$ = of(true);
+  readyAuthentication$ = of(default_headers);
+}
+
+class authServiceStub_failed {
+  isLoggedIn$ = of(false);
+  readyAuthentication$ = throwError("Can't authenticate: no credentials");
+}
+
+describe("ApiService_failed_auth", () => {
   beforeEach(() => {
-    TestBed.configureTestingModule({
-      providers: [ApiService, AuthService],
-      imports: [HttpClientTestingModule]
-    });
-
-    httpMock = TestBed.get(HttpTestingController as Type<HttpTestingController>);
-    service = TestBed.get(ApiService as Type<ApiService>);
-    auth = TestBed.get(AuthService as Type<AuthService>);
-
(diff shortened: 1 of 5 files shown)


Drvanon:
Hahaha, all ways pointing to master. Happens to me all the time as well. Will look at this later.

gridt/src/app/core/api.service.ts
```diff
@@ -225,4 +231,42 @@ export class ApiService {
       catchError( this.handleBadAuth() )
     );
   }
+
+   /**
+    * Observable to obtain identity from server
+    */
+
+  public Identity$(): Observable<Identity> {
```
Drvanon: Because it is an instance of identity, generally these things are set to lower case.

Drvanon: Also this is a function that takes no input, and returns one specific value, it might be more sensible to make this just a property.

gridt/src/app/core/api.service.ts
```diff
@@ -225,4 +231,42 @@ export class ApiService {
...
+      flatMap((options) => this.http.get<Identity>(
+        `${this.URL}/identity`,
+        options
+      )),
+      catchError( this.handleBadAuth())
+    )
+  }
+
+  public changeBio$( bio: string ) {
+   	console.debug(`Saving new biography to the server. (at least it should now create a http.put)`);
```
Drvanon: The indentation is not aligned here.

Drvanon: Also please remove that thing between the parenthesis, it is not useful information.

gridt/src/app/core/api.service.ts
```diff
@@ -225,4 +231,42 @@ export class ApiService {
...
+
+    return this.auth.readyAuthentication$.pipe(
+     flatMap((options) => this.http.put(
```
Drvanon: Inconsistent indentation.

gridt/src/app/core/api.service.ts
```diff
@@ -225,4 +231,42 @@ export class ApiService {
       catchError( this.handleBadAuth() )
     );
   }
+
+   /**
+    * Observable to obtain identity from server
+    */
+
```
Drvanon: Unnecessary white space.

gridt/src/app/core/api.service.ts
```diff
@@ -1,12 +1,18 @@
...
 import { Movement } from "./models/movement.model";
 import { User } from './models/user.model';
-import { ServerMessage } from './models/server-responses.model';
+import { Identity } from './models/identity.model';
+
 import { AuthService } from './auth.service';
 
+export interface ServerMessage {
+  message: string;
+}
```
Drvanon: Why did you redo this?

gridt/src/app/core/api.service.spec.ts
```diff
@@ -86,6 +88,17 @@ describe("ApiService", () => {
...
+
+    mock_identity = [
+      {
+        id: 1,
+        username: "ALittleOne",
+        bio: "Less then one foot tall.",
+        email: "a_little@one.com",
+        avatar: "arandomstring"
+      }
+    ]
```
Drvanon: Please end with a semicolon.

Drvanon: Also it is good practice to declare the interfaces it implements.

gridt/src/app/core/api.service.spec.ts
```diff
@@ -253,6 +266,72 @@ describe("ApiService", () => {
...
+      service.Identity$()
+    ).toBeObservable(cold("#", {}, "Can't authenticate: no credentials"));
   });
+
+  it('should fail to update bio when not logged in', () => {
+    authServiceStub.readyAuthentication$ = cold('#', null , "Can't authenticate: no credentials");
+    service = new ApiService(httpClientStub, authServiceStub);
+    let mock_bio = "My very first bio."
+    expect(service.changeBio$( mock_bio )).toBeTruthy( cold('#', null, "Can't authenticate: no credentials"))
+  })
```
Drvanon: Don't forget the semicolons.

gridt/src/app/core/api.service.spec.ts
```diff
@@ -253,6 +266,72 @@ describe("ApiService", () => {
...
-    )
+    );
+  });
+
+  it('should fail to read identity from server when not logged in', () => {
+    authServiceStub.readyAuthentication$ = cold('#', null , "Can't authenticate: no credentials");
+    service = new ApiService(httpClientStub, authServiceStub);
+
+    expect(
+      service.Identity$()
```
Drvanon: Semicolon.

gridt/src/app/core/api.service.spec.ts
```diff
@@ -253,6 +266,72 @@ describe("ApiService", () => {
...
+  })
+
+  it('should be able to update the user bio', () => {
+    httpClientStub.put.and.returnValue(of(
+      { "message": "Succesfully updated bio." }
+    ));
+
+    service = new ApiService(httpClientStub, authServiceStub);
+
+    let mock_bio = "My insanely good new bio."
```
Drvanon: ;

gridt/src/app/core/api.service.spec.ts
```diff
@@ -253,6 +266,72 @@ describe("ApiService", () => {
...
+      default_headers
+    );
+  })
+
+  it('should be able to update the user bio', () => {
+    httpClientStub.put.and.returnValue(of(
+      { "message": "Succesfully updated bio." }
+    ));
+
+    service = new ApiService(httpClientStub, authServiceStub);
```
Drvanon: Please use dependency injection, I know this can be a quite the pain, but I think it is not honest to circumvent it like this, because there could be things going wrong in this section of the code (e.g. incorrectly configured modules) that would not be caught this way. You can create a new test suite or (and I think this is more beautiful) reset the TestBed using `TestBed.resetTestingModule(); ` that does not solve everything, but it helps.

gridt/src/app/core/api.service.spec.ts
```diff
@@ -253,6 +266,72 @@ describe("ApiService", () => {
...
+
+    expect(
+      service.Identity$()
+    ).toBeObservable(cold("#", {}, "Can't authenticate: no credentials"));
   });
+
+  it('should fail to update bio when not logged in', () => {
+    authServiceStub.readyAuthentication$ = cold('#', null , "Can't authenticate: no credentials");
+    service = new ApiService(httpClientStub, authServiceStub);
```
Drvanon: Please see the comment on dependency injection.

Drvanon (changes requested):
I would really appreciate it if you could add some comments to the newly created properties and functions in the style of sphinx. I tried to do this previously but it has enough value to push for this I think.

gridt/src/app/core/api.service.spec.ts
marickmanrho: You mean what I do in line 18?

marickmanrho:
It is a mystery to me why the test is passing now... We have to get this sorted asap.

gridt/src/app/core/models/identity.model.ts
```diff
@@ -0,0 +1,22 @@
+export interface Identity {
```
Drvanon, line 1: Does it make sense to make this a [partial implementation](https://stackoverflow.com/a/57070274/1051438) of User?

gridt/src/app/movements/add-movement/add-movement.page.spec.ts
```diff
@@ -1,6 +1,6 @@
 import { of } from 'rxjs';
```
Drvanon: This theoretically does not belong in this request.

gridt/src/app/movements/movements-detail/movements-detail.page.spec.ts
```diff
@@ -1,14 +1,21 @@
 import { CUSTOM_ELEMENTS_SCHEMA } from "@angular/core";
```
Drvanon: This technically is outside the scope of the request.

gridt/src/app/core/models/identity.model.ts
marickmanrho: It does make sense. Although, if I want to do that I want to exclude the last_signal part. The Identity should be complete for new users and should not require some first interactions with a movement. Hence, in order to do this the User model should be split also. I don't think that is something we want to change now. Maybe an audit of the models in general can be beneficial to the code quality when the settings page is done.

gridt/src/app/core/api.service.spec.ts
```diff
@@ -1,96 +1,155 @@
...
-
+let mock_identity: Identity[] = [
+  {
+    id: 1,
+    username: "ALittleOne",
+    bio: "Less then one foot tall.",
+    email: "a_little@one.com",
+    avatar: "arandomstring"
+  }
+];
```
Drvanon: This is a list of size 1. Add more identities or make it a single.

gridt/src/app/core/api.service.spec.ts
```diff
@@ -1,96 +1,155 @@
...
-        interval: "daily"
-      }
-    ];
-    mock_subscriptions = [...mock_movements].filter(m => m.subscribed);
-  });
+    service = TestBed.get(ApiService as Type<ApiService>);
 
-  afterEach(() => {
-    httpMock.verify();
+    mock_subscriptions = [...mock_movements].filter(m => m.subscribed);
```
Drvanon, line 166: This is an inititialization for the next round, it is more logical to do this in `beforeEach` and remove this function.

gridt/src/app/core/api.service.spec.ts
```diff
@@ -253,6 +294,37 @@ describe("ApiService", () => {
       `${service.URL}/movements/${movement_id}/leader/1`,
       {},
       default_headers
-    )
+    );
+  });
+
+  it('should be able to retreive identity', () => {
+    httpClientStub.get.and.returnValue(of(mock_identity[0]));
```
Drvanon, line 315: Theoretically you could also confirm that the right value is URL is called by using `expect(httpClientStub.get).toBeCalledWith({some_object})`. You could do this for the PUT requests as well, if you think it is worth it.

Drvanon (commented):
Good work! Please look at my comments and consider them.

gridt/src/app/core/api.service.spec.ts
marickmanrho: But it is in beforeEach right now, and indeed the afterEach() function is removed.

marickmanrho:
Due to reverting the changes that where outside the scope of this project the test is now failing. My changes do not cause any errors.

Edit: See also PR #158 which fixes the tests.

Drvanon (approved):
You can feel free to merge it. Can't do that from phone.

Issue #138, opened 2020-09-08
As a community manager I would like that pull requests to development would be tested immediately so that I can know if I need to start reviewing.


Drvanon:
Pull requests always point to api-development in our project, but it is excluded from the CI. Usually I do not need to pay a lot of attention to branches whose tests fail, so it is really useful for me to know.

HendrikPreuter:
This needs an update as we no longer point to api-development.

marickmanrho:
I think this is fixed, hence I close this issue.

PR #144, opened 2020-09-09
Adds Funding and CI implementation to development.

marickmanrho:
Unfortunately I think this is needed to make the CI work on the development branch. For example look at PR #143 which now doesn't have tests being run automatically.

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,3 @@
+# These are supported funding model platforms
+
+open_collective: gridt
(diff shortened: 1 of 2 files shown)


marickmanrho:
Fixes: #138

Drvanon:
I am very confused: why are there two CI's? And why is one of them failing?

marickmanrho:
When I try to reproduce this error on my device it is not always showing up. That might be the reason here to.

marickmanrho:
Nevertheless I think we should merge this pull request. Then try to fix this by solving Issue #146

Drvanon:
Noted.

PR #158, opened 2020-09-10
Fix authStub.

marickmanrho:
Adds:

Correct authentication stubbing on add-movement.page.spec.ts and movements-details.page.spec.ts.


Drvanon:
Looks good, if you merge it now some of your merge requests will stop erroring.

Issue #157, opened 2020-09-10
Authentication Stub is not implemented correctly on movement subpages.


marickmanrho:
On both add-movement.page.spec.ts and movements-detail.page.spec.ts the authentication stub is not defined properly causing the tests to fail.

PR #159, opened 2020-09-10
Store user identity in localstorage

### LINKED ISSUE #148: As a Client, I want to store my Identity locally, so I can review them when I have no internet connection.
The identity of the user should be stored locally in a secure storage. The users can then retrieve their identity regardless of their internet connection.

marickmanrho:
Adds:

- A new setting service which is able to store and retrieve the user's identity locally.

fixes #148 

Remarks:
There may be a nicer way to define `setLocalIdentity$(...)`. I coded it this way because:
- I want to avoid having to subscribe to the returned observable in order to actually store information into the localstorage. 
- Still want it to give a proper error when user is not authenticated.
- Avoid having uncatched errors produced by, for example, `readyAuthentication$`.
Suggestions for improvement are welcome.

### DIFF
--- a/gridt/src/app/core/api.service.spec.ts
+++ b/gridt/src/app/core/api.service.spec.ts
@@ -96,7 +96,7 @@ class authServiceStub_failed {
   readyAuthentication$ = throwError("Can't authenticate: no credentials");
 }
 
-describe("ApiService_failed_auth", () => {
+describe("ApiService when authentication fails", () => {
   beforeEach(() => {
     httpClientStub = jasmine.createSpyObj(
       'httpClient', ['get', 'post', 'put']
@@ -135,18 +135,18 @@ describe("ApiService_failed_auth", () => {
 
   it('should fail to update bio when not logged in', () => {
     let mock_bio = "My very first bio."
-    expect(service.changeBio$( mock_bio )).toBeTruthy( cold('#', null, "Can't authenticate: no credentials"));
+    expect(service.changeBio$( mock_bio )).toBeObservable( cold('#', null, "Can't authenticate: no credentials"));
   });
 
   it('should fail to update password when not logged in', () => {
     let old_password = "ABCDEF";
     let new_password = "abcdef";
 
-    expect(service.changePassword$( old_password, new_password )).toBeTruthy( cold('#', null, "Can't authenticate: no credentials"));
+    expect(service.changePassword$( old_password, new_password )).toBeObservable( cold('#', null, "Can't authenticate: no credentials"));
   });
 });
 
-describe("ApiService_succesful_auth", () => {
+describe("ApiService when authentication is succesful", () => {
 
   beforeEach(() => {
     httpClientStub = jasmine.createSpyObj(
@@ -314,7 +314,7 @@ describe("ApiService_succesful_auth", () => {
   it('should be able to retreive identity', () => {
     httpClientStub.get.and.returnValue(of(mock_identity[0]));
 
-    expect(service.userIdentity$).toBeObservable(cold('(a|)',{a: mock_identity[0]}));
+    expect(service.userIdentity$).toBeObservable(cold('(a|)', {a: mock_identity[0]}));
     expect(httpClientStub.get).toHaveBeenCalledWith(
       `${service.URL}/identity`,
       default_headers
--- a/gridt/src/app/core/api.service.spec.ts
+++ b/gridt/src/app/core/api.service.spec.ts
@@ -115,7 +115,7 @@ describe("ApiService_failed_auth", () => {
 
   it('should fail to read identity from server when not logged in', () => {
     expect(
-      service.userIdentity$()
+      service.userIdentity$
     ).toBeObservable(cold("#", {}, "Can't authenticate: no credentials"));
   });
 
@@ -300,7 +300,7 @@ describe("ApiService_succesful_auth", () => {
   it('should be able to retreive identity', () => {
     httpClientStub.get.and.returnValue(of(mock_identity[0]));
 
-    expect(service.userIdentity$()).toBeObservable(cold('(a|)',{a: mock_identity[0]}));
+    expect(service.userIdentity$).toBeObservable(cold('(a|)',{a: mock_identity[0]}));
   });
 
   it('should be able to update the user bio', () => {
--- a/gridt/src/app/core/api.service.spec.ts
+++ b/gridt/src/app/core/api.service.spec.ts
@@ -1,109 +1,155 @@
-import {Type } from "@angular/core";
+import { Type } from "@angular/core";
 import { TestBed } from "@angular/core/testing";
+import { HttpClient, HttpHeaders, HttpErrorResponse } from '@angular/common/http';
 import { HttpClientTestingModule, HttpTestingController } from "@angular/common/http/testing";
 
-import { forkJoin, of, throwError } from "rxjs";
-import { skip, take, toArray } from "rxjs/operators";
+import { cold } from 'jasmine-marbles';
+import { of, throwError } from "rxjs";
 
+import { AuthService } from './auth.service';
 import { ApiService } from "./api.service";
+
 import { Movement } from "./models/movement.model";
 import { Identity } from "./models/identity.model";
 import { User } from './models/user.model';
-import { AuthService } from './auth.service';
 
-import { cold } from 'jasmine-marbles';
-import { HttpClient, HttpHeaders, HttpResponse, HttpHeaderResponse, HttpErrorResponse } from '@angular/common/http';
+let mock_movements: Movement[] = [
+  {
+    id: 10129312983,
+    name: "Mock Name",
+    short_description: "Mocking the short description",
+    description: "I will mock description. You are a fool, description.",
+    subscribed: false,
+    interval: "daily"
+  },
+  {
+    id: 1283912983123,
+    name: "Another mocked movement",
+    short_description: "This one only has a short description and no long one",
+    subscribed: false,
+    interval: "twice daily"
+  },
+  {
+    id: 10023123,
+    name: "Subscribed movement",
+    short_description: "This movement is supposed to come from the server.",
+    description: "We need some way to fake a response from the server, telling the client that it \
+     is subscribed to this particular movement",
+    subscribed: true,
+    interval: "daily"
+  },
+  {
+    id: 112312983,
+    name: "Flossing",
+    short_description: "Floss every day",
+    description: "We floss every day because it is good for our theeth.",
+    subscribed: true,
+    interval: "daily"
+  }
+];
 
-let mock_movements: Movement[];
 let mock_subscriptions: Movement[];
-let mock_identity: Identity[];
 
-describe("ApiService", () => {
-  let service: ApiService;
-  let auth: AuthService;
-  let httpMock: HttpTestingController;
-  let httpClientStub: jasmine.SpyObj<HttpClient>;
-  let authServiceStub: jasmine.SpyObj<AuthService>;
+let mock_identity: Identity[] = [
+  {
+    id: 1,
+    username: "ALittleOne",
+    bio: "Less then one foot tall.",
+    email: "a_little@one.com",
+    avatar: "arandomstring"
+  }
+];
 
-  const default_headers = {
-    headers: new HttpHeaders({
-      Authorization: "JWT aksdajskd.asdjknaskdn.asdjknakdnasjd"
-    })
-  };
+const default_headers = {
+  headers: new HttpHeaders({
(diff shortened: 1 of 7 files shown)


marickmanrho:
Please merge PR #157 , #143 first as this PR already includes them!

Drvanon:
#157 is not implemented yet. Should I be reviewing this right now?

marickmanrho:
Yes. My mistake, that issue should have been closed together with PR #158

gridt/src/app/core/api.service.spec.ts
```diff
@@ -1,96 +1,169 @@
...
+  isLoggedIn$ = of(true);
+  readyAuthentication$ = of(default_headers);
+}
+
+class authServiceStub_failed {
+  isLoggedIn$ = of(false);
+  readyAuthentication$ = throwError("Can't authenticate: no credentials");
+}
+
+describe("ApiService_failed_auth", () => {
```
Drvanon: This is a normal string, so you can make it a nice statement like `"ApiService when auth fails"`, so that the final sentence would become something like "ApiService when auth fails should create".

gridt/src/app/core/api.service.spec.ts
```diff
@@ -1,96 +1,169 @@
...
+
+  it('should fail to read identity from server when not logged in', () => {
+    expect(
+      service.userIdentity$
+    ).toBeObservable(cold("#", {}, "Can't authenticate: no credentials"));
+  });
+
+  it('should fail to update bio when not logged in', () => {
+    let mock_bio = "My very first bio."
+    expect(service.changeBio$( mock_bio )).toBeTruthy( cold('#', null, "Can't authenticate: no credentials"));
```
Drvanon: This is the wrong jasmine call.

gridt/src/app/core/api.service.spec.ts
```diff
@@ -1,96 +1,169 @@
...
+  });
+
+  it('should fail to update password when not logged in', () => {
+    let old_password = "ABCDEF";
+    let new_password = "abcdef";
+
+    expect(service.changePassword$( old_password, new_password )).toBeTruthy( cold('#', null, "Can't authenticate: no credentials"));
```
Drvanon: Again, this should be `toBeObservable`.

gridt/src/app/core/api.service.spec.ts
```diff
@@ -253,6 +308,47 @@ describe("ApiService", () => {
...
       {},
       default_headers
-    )
+    );
+  });
+
+  it('should be able to retreive identity', () => {
+    httpClientStub.get.and.returnValue(of(mock_identity[0]));
+
+    expect(service.userIdentity$).toBeObservable(cold('(a|)',{a: mock_identity[0]}));
```
Drvanon: Space after the comma is nice to have.

gridt/src/app/core/api.service.ts
```diff
@@ -79,15 +83,15 @@ export class ApiService {
    * @param movement The new version of the movement.
    */
   private replace_movement_in_bsubject(bsubject: BehaviorSubject<Movement[]>, movement: Movement) {
```
Drvanon: Can't wait to remove this function.

gridt/src/app/core/api.service.ts
```diff
@@ -225,4 +229,48 @@ export class ApiService {
...
+    )),
+    catchError( this.handleBadAuth())
+  );
+
+  /**
+  * Changes the biography of the user on the server.
+  * @param bio the new bio that the user wants.
+  */
+  public changeBio$( bio: string ) {
+   	console.debug(`Saving new biography to the server.`);
```
Drvanon: Inconsistent indentation.

gridt/src/app/core/settings.service.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+
+class authServiceStub_succes {
+  isLoggedIn$ = of(true);
+  readyAuthentication$ = of(default_headers);
+};
+
+class authServiceStub_failed {
+  isLoggedIn$ = of(false);
+  readyAuthentication$ = throwError("Can't authenticate: no credentials");
+  error_codes = {NOCREDENTIALS: "Can't authenticate: no credentials"};
```
Drvanon: Why is this necessary?

gridt/src/app/movements/movements-detail/movements-detail.page.spec.ts
```diff
@@ -1,14 +1,21 @@
 import { CUSTOM_ELEMENTS_SCHEMA } from "@angular/core";
 import { async, ComponentFixture, TestBed } from "@angular/core/testing";
 import { RouterTestingModule } from "@angular/router/testing";
-import { HttpClientModule } from "@angular/common/http";
+import { HttpClientModule, HttpHeaders } from "@angular/common/http";
```
Drvanon: Are the changes in this file intentional?

gridt/src/app/core/settings.service.ts
```diff
@@ -0,0 +1,44 @@
...
+  );
+
+  /**
+   * Store settings in the secure storage.
+   * @param identity Settings to be stored.
+   */
+  public setLocalIdentity$(identity: Identity): Observable<boolean> {
+    console.debug(`Storing identity: ${JSON.stringify(identity)} into the local storage.`);
+    let loggedIn: boolean;
+    this.auth.isLoggedIn$.subscribe( (val) => {loggedIn = val;});
```
Drvanon: You should deal with this subscription and make sure it does not create a memory leak. @hendrikpreuter has become an expert in this, I would discuss the matter with him.

gridt/src/app/core/settings.service.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    if (loggedIn) {
```
Drvanon: This should always fail when you are online. Your test should make the wait time a little longer so that you can see that in your tests, but the subscription will only be called once the request responds. That might be a few cycles later, meaning that this will be evaluated to false immediately. The logic you have placed here should be in the subscription function.

Drvanon (changes requested):
The comments in settings.service.ts will have to be addressed before I can approve. The rest are very minor changes.

gridt/src/app/core/settings.service.ts
```diff
@@ -30,15 +30,15 @@ export class SettingsService {
...
-    if (loggedIn) {
-      this.secStore.set$("identity", identity).subscribe();
-      return of(true);
-    } else {
-      return throwError("Not logged in: can't store identity in local storage.");
-    }
+    return this.auth.isLoggedIn$.pipe(
+      take(1),
+      tap( (val) => {
+        if (val === true) {
```
Drvanon: This should be a filter 

https://www.learnrxjs.io/learn-rxjs/operators/filtering/filter

gridt/src/app/core/settings.service.ts
```diff
@@ -0,0 +1,44 @@
...
+   * Store settings in the secure storage.
+   * @param identity Settings to be stored.
+   */
+  public setLocalIdentity$(identity: Identity): Observable<boolean> {
+    return this.auth.isLoggedIn$.pipe(
+      take(1),
+      tap( (val) => {
+        if (val === true) {
+          console.debug(`Storing identity: ${JSON.stringify(identity)} into the local storage.`);
+          this.secStore.set$("identity", identity).subscribe();
```
Drvanon: This should be a flat map

https://www.learnrxjs.io/learn-rxjs/operators/transformation/mergemap

gridt/src/app/core/settings.service.ts
```diff
@@ -0,0 +1,44 @@
...
+        }
+      }),
+      mergeMap( val => iif( () => val === true, of(true), throwError("Not logged in: can't store identity in local storage.")))
```
Drvanon: This should be a mapTo

https://www.learnrxjs.io/learn-rxjs/operators/transformation/mapto

And a catchError/throwError thing

Drvanon (changes requested):
I am seeing the logic I want to see, but not in a manner I deem maintainable.

gridt/src/app/core/settings.service.ts
marickmanrho: I'm not sure about that. If you use a filter as you say then upon a failed auth, it will just do nothing at all. While I would like it to then throw an error. Hence I used tap, such that any value will be passed to the map.

gridt/src/app/core/settings.service.ts
marickmanrho: They are the same. The website says mergeMap is an alias for flatMap. So why is this wrong?

marickmanrho: Ooh, sorry I was looking at the wrong place for this comment.

marickmanrho: Why? I don't want the output of secStore.set$ to be the new observable of setLocalIdentity$.

gridt/src/app/core/settings.service.ts
marickmanrho: I don't see why you'd want this to map to a single value. The whole idea of using the `iif()` is to return `of(true)` when successful and and error when not.

edit: iif

Drvanon: If there is an error, mapTo should ignore it and just pass on the error. The error you emit is different from the one you get from secstore so a catchError/throwError can be used to perform that replacement.

From the filter you know you will only obtain values that are true, so this observable will always emit true, unless it emits an observable, mapTo makes that very clear.

gridt/src/app/core/settings.service.ts
Drvanon: The output of isLoggedIn is a Boolean and after observing that you are, this value is no longer important. Once that step is completed, the only thing we are still interested in, is if the settings were stored properly. If they were we can return true. If not then we throw an error. Now the error catching will have to be done in a way that it is able to distinguish between an error caused by a failed login and a failed setting, but all of this is clearly very descriptive in the sense that rxjs is useful for describing how it should be done. 


Note: in the current state you don't do anything with the case that the user is not loggedIn, it would be good to throw an error when he is not loggedIn so that for the observer it is clear that it failed and why.

gridt/src/app/core/settings.service.ts
Drvanon: Fair, throwing away the error would not be a good idea. You could use flatmap to throw an error on the false condition and true on the other.

gridt/src/app/core/settings.service.ts
```diff
@@ -0,0 +1,44 @@
...
+    )
+  }
```
Drvanon: This is what I would recommend to put in the pipe:
```ts
this.auth.isLoggedIn$.pipe(
  flatMap( (loggedIn) => {
    const setter = this.secStore.set$("identity", identity).pipe(
       catchError(() => throwError("Could not set identity: secure storage not available".))
    );
    return (loggedIn) ? setter: throwError("Could not set identity: not logged in");
  }), 
);
```
This has the following behaviour:
- If the user is not logged in, it throws an error.
- If the user is logged in, but the secure storage fails to set the identity, it throws an error.
- If the user is logged in and the secure storage sets the identity successfully, it will return the output of `secureStorage.set$`. (This "nexts" `true` in case the setting was successful.)

Drvanon: If you'd go for this, I would recommend maintaining a error codes variable and drawing from there, but this was more for the idea.

marickmanrho: Great, I took the liberty to alter the code a little bit. I decided to use `readyAuthentication$` such that `secStore.set$` is never called when not logged in.

gridt/src/app/core/settings.service.ts
```diff
@@ -30,15 +30,13 @@ export class SettingsService {
...
-      }),
-      mergeMap( val => iif( () => val === true, of(true), throwError("Not logged in: can't store identity in local storage.")))
+    return this.auth.readyAuthentication$.pipe(
+      catchError( () => throwError("Could not set identity: not logged in")),
+      take(1), // Makes sure the observable completes
+      flatMap( () => this.secStore.set$("identity", identity).pipe(
+          catchError(()=> throwError("Could not set identity: secure storage not available."))
+        )
+      )
     )
```
Drvanon: There should be a semi colon here and maybe you can use the error code pattern here.

These are the last pieces of correction I'd like to give before accepting and merging the code.

marickmanrho: I took the liberty to also add some custom errors for the localIdentity$ observable. The error codes are also used in the test, such that we don't have to change them twice.

Drvanon (approved):
Unless you feel like something should change, I say, let's merge.

Side note: generally speaking the error code is just straight up an error code and not a composition of two. You can leave it like this or change it, the quality of the code is changed only in a very minor way.

PR #174, opened 2020-09-23
135 subscription audit

### LINKED ISSUE #135: The application should be audited for unsubscriptions
RxJs can create memory leaks by subscribing to an observable and never completing it. (https://itnext.io/angular-rxjs-detecting-memory-leaks-bdd312a070a0)

It would be good to search the entire application for "subscribe" and checking if the observable is guaranteed to either complete, error or unsubscribe.

HendrikPreuter:
#135 

To-do:
-Awaiting round 2 of feedback

Done:
- added unsubscribe to all subscriptions via an extendable class
- added tests for all these unsubscribe cases
- added timer for secStorage
-add throw error in empty subscriptions

### DIFF
--- a/gridt/src/app/app.component.ts
+++ b/gridt/src/app/app.component.ts
@@ -64,7 +64,7 @@ export class AppComponent implements OnInit, OnDestroy {
       this.checkAuthOnResume.bind(this)
     );
   }
-  
+
   ngOnDestroy() {
     this.cancelAllSubscriptions();
   }
--- a/gridt/src/app/app.component.ts
+++ b/gridt/src/app/app.component.ts
@@ -11,9 +11,9 @@ import { filter, flatMap } from 'rxjs/operators';
   selector: 'app-root',
   templateUrl: 'app.component.html'
 })
-export class AppComponent implements OnInit {
+export class AppComponent implements OnInit, OnDestroy {
   public isLoggedIn$: Observable<boolean>;
-  private subscriptions: Subscription;
+  private isLoggedInSubscription: Subscription;
 
   public appPages = [
     {
@@ -64,15 +64,21 @@ export class AppComponent implements OnInit {
       this.checkAuthOnResume.bind(this)
     );
   }
+  
+  ngOnDestroy() {
+    this.cancelAllSubscriptions();
+  }
 
-  public unsubscribeLoggedIn() {
-    this.subscriptions.unsubscribe();
+  private cancelAllSubscriptions() {
+    if (this.isLoggedInSubscription) {
+      this.isLoggedInSubscription.unsubscribe();
+    }
   }
 
   private checkAuthOnResume(state: AppState) {
     this.zone.run( () => {
       if (state.isActive) {
-        this.subscriptions = this.auth.isLoggedIn$.subscribe(loggedIn => {
+        this.isLoggedInSubscription = this.auth.isLoggedIn$.subscribe(loggedIn => {
           if (!loggedIn) {
             console.log('On resuming the app found user is now logged out. Redirecting to /login');
             this.router.navigate(['/login']);
--- a/gridt/src/app/app.component.ts
+++ b/gridt/src/app/app.component.ts
@@ -13,7 +13,6 @@ import { filter, flatMap } from 'rxjs/operators';
 })
 export class AppComponent implements OnInit {
   public isLoggedIn$: Observable<boolean>;
-  private subscriptions: Subscription[] = [];
 
   public appPages = [
     {
@@ -65,19 +64,15 @@ export class AppComponent implements OnInit {
     );
   }
 
-  private cancelAllSubscriptions() {
-    this.subscriptions.map( subscription => subscription.unsubscribe());
-  }
-
   private checkAuthOnResume(state: AppState) {
     this.zone.run( () => {
       if (state.isActive) {
-        this.subscriptions.push(this.auth.isLoggedIn$.subscribe(loggedIn => {
+        this.auth.isLoggedIn$.subscribe(loggedIn => {
           if (!loggedIn) {
             console.log('On resuming the app found user is now logged out. Redirecting to /login');
             this.router.navigate(['/login']);
           }
-        }));
+        });
       }
     });
   }
--- a/gridt/src/app/app.component.ts
+++ b/gridt/src/app/app.component.ts
@@ -4,15 +4,16 @@ import { Router, NavigationStart } from '@angular/router';
 import { SplashScreen } from '@ionic-native/splash-screen/ngx';
 import { Plugins, AppState, Capacitor } from '@capacitor/core';
 import { AuthService } from './core/auth.service';
-import { Observable } from 'rxjs';
+import { Observable, Subscription } from 'rxjs';
 import { filter, flatMap } from 'rxjs/operators';
 
 @Component({
   selector: 'app-root',
   templateUrl: 'app.component.html'
 })
-export class AppComponent implements OnInit, OnDestroy{
+export class AppComponent implements OnInit {
   public isLoggedIn$: Observable<boolean>;
+  private subscriptions: Subscription;
 
   public appPages = [
     {
@@ -64,15 +65,14 @@ export class AppComponent implements OnInit, OnDestroy{
     );
   }
 
-  onLogout() {
+  public unsubscribeLoggedIn() {
+    this.subscriptions.unsubscribe();
   }
-
-  ngOnDestroy() { }
 
   private checkAuthOnResume(state: AppState) {
     this.zone.run( () => {
       if (state.isActive) {
-        this.auth.isLoggedIn$.subscribe(loggedIn => {
+        this.subscriptions = this.auth.isLoggedIn$.subscribe(loggedIn => {
           if (!loggedIn) {
             console.log('On resuming the app found user is now logged out. Redirecting to /login');
             this.router.navigate(['/login']);
--- a/gridt/src/app/app.component.ts
+++ b/gridt/src/app/app.component.ts
@@ -11,9 +11,9 @@ import { filter, flatMap } from 'rxjs/operators';
   selector: 'app-root',
   templateUrl: 'app.component.html'
 })
-export class AppComponent implements OnInit, OnDestroy {
+export class AppComponent implements OnInit {
   public isLoggedIn$: Observable<boolean>;
-  private isLoggedInSubscription: Subscription;
+  private subscriptions: Subscription[] = [];
 
   public appPages = [
     {
@@ -65,25 +65,19 @@ export class AppComponent implements OnInit, OnDestroy {
     );
   }
 
-  ngOnDestroy() {
-    this.cancelAllSubscriptions();
-  }
-
   private cancelAllSubscriptions() {
-    if (this.isLoggedInSubscription) {
-      this.isLoggedInSubscription.unsubscribe();
-    }
+    this.subscriptions.map( subscription => subscription.unsubscribe());
   }
 
   private checkAuthOnResume(state: AppState) {
     this.zone.run( () => {
       if (state.isActive) {
-        this.isLoggedInSubscription = this.auth.isLoggedIn$.subscribe(loggedIn => {
+        this.subscriptions.push(this.auth.isLoggedIn$.subscribe(loggedIn => {
           if (!loggedIn) {
             console.log('On resuming the app found user is now logged out. Redirecting to /login');
             this.router.navigate(['/login']);
           }
-        });
+        }));
       }
     });
   }
--- a/gridt/src/app/core/api.service.ts
+++ b/gridt/src/app/core/api.service.ts
@@ -15,17 +15,9 @@ import { ServerMessage } from "./models/server-responses.model";
   providedIn: "root"
 })
 export class ApiService {
-  get allMovements$ (): Observable<Movement[]> {
-    return this._allMovements$.asObservable();
-  }
-  get subscriptions$ (): Observable<Movement[]> {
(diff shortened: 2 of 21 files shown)


Drvanon (changes requested):
I notice you tend to go for registering individual subscriptions and unsubcribing from them. Because the principle of unsubscribing is generally applicable and there is no further use for the subscriptions, it makes sense to gather them into one list and remove that list at the onDestroy.

Drvanon:
Also, on the services and app.component, onDestroy will never be called. Sorry for not mentioning this earlier, but the page will just be destroyed. There are places where it might make sense to put time outs, if a subscription is performed in a function and then never looked after again. If the time out happens, it should however throw a publicly visible error.

HendrikPreuter:
> I notice you tend to go for registering individual subscriptions and unsubcribing from them. Because the principle of unsubscribing is generally applicable and there is no further use for the subscriptions, it makes sense to gather them into one list and remove that list at the onDestroy.

Does that mean i still need to unsubscribe and then empty the list? Or just empty the list of subscriptions, because i feel that doesn't end them rightly.

> Also, on the services and app.component, onDestroy will never be called. Sorry for not mentioning this earlier, but the page will just be destroyed. There are places where it might make sense to put time outs, if a subscription is performed in a function and then never looked after again. If the time out happens, it should however throw a publicly visible error.

We had a conversation about this and i remembered the part where your said a service gets terminated with the component it is attached to. However i forgot the only component they are attached to is in fact app.component. We also somewhat talked about these timeouts. You wanted something in sec storage right? WIth the takeuntil(5) or something along those lines. However i might still need some more training in rxjs for this as i only found 1 of the functions where i could put the .pipe(takeUntil(timer(5000))). And i was not even sure that one was the right spot.

Is what you are suggesting here something else or is it what i mentioned above?

Drvanon:
The most important thing is that you unsubscribe onDestroy. The garbage collector can then itself determine what it will do with the subscription after the component is destroyed.

`.pipe(takeUntil(timer(5000))).` is indeed what you are looking for. I don't know if 5 seconds is the right time (haven't given it much thought), but that would make sure it doesn't crash. However, it would be good for debugging purposes, if you throw an error outside of the pipe (aka in the subscription), because nothing should fail silently.

gridt/src/app/app.component.ts
```diff
@@ -65,25 +65,19 @@ export class AppComponent implements OnInit, OnDestroy {
...
-      this.isLoggedInSubscription.unsubscribe();
-    }
+    this.subscriptions.map( subscription => subscription.unsubscribe());
   }
 
   private checkAuthOnResume(state: AppState) {
     this.zone.run( () => {
       if (state.isActive) {
-        this.isLoggedInSubscription = this.auth.isLoggedIn$.subscribe(loggedIn => {
+        this.subscriptions.push(this.auth.isLoggedIn$.subscribe(loggedIn => {
```
Drvanon: Because this takes place in the `app.component` there is no need to deal with the subscribe/unsubscribe logic.

gridt/src/app/core/secure-storage.service.ts
```diff
@@ -1,14 +1,19 @@
...
+import { takeUntil } from 'rxjs/operators';
 
 const { SecureStoragePlugin } = Plugins;
 
 @Injectable({
   providedIn: 'root'
 })
 export class SecureStorageService {
+  // setting the timer for the observables to finnish
+  private seconds = 5000;
```
Drvanon: This should be more descriptive.

gridt/src/app/core/secure-storage.service.ts
```diff
@@ -1,14 +1,19 @@
...
 import { Injectable } from '@angular/core';
+import { takeUntil } from 'rxjs/operators';
 
 const { SecureStoragePlugin } = Plugins;
 
 @Injectable({
   providedIn: 'root'
 })
 export class SecureStorageService {
+  // setting the timer for the observables to finnish
```
Drvanon: Spelling error in finish

gridt/src/app/core/secure-storage.service.ts
```diff
@@ -23,23 +28,24 @@ export class SecureStorageService {
           }
         }
       ).catch( () => observer.error(`Key "${key}" does not exist in the secure storage.`) );
-    });
+    }).pipe(takeUntil(this.timer));
```
Drvanon, line 32: Please leave a comment as to why this is necessary. (We know now, but for future reference.)

gridt/src/app/core/secure-storage.service.ts
```diff
@@ -23,23 +28,24 @@ export class SecureStorageService {
...
   }
 
   set$ (key: string, value: any): Observable<boolean> {
-    return new Observable ( (observer) => {
+    return new Observable<boolean> ( (observer) => {
       SecureStoragePlugin.set({key, value}).then( (succesObj) => {
         if (succesObj.value) {
           observer.next(true);
+          observer.complete();
```
Drvanon, line 40: Good idea.

gridt/src/app/home/home.page.ts
```diff
@@ -35,15 +33,7 @@ export class HomePage implements OnInit, OnDestroy {
...
-    if (this.swapLeaderSubscription) {
-      this.swapLeaderSubscription.unsubscribe();
-    }
-    if (this.sendSignalSubscription) {
-      this.sendSignalSubscription.unsubscribe();
-    }
-    if (this.getMovementsSubscription) {
-      this.getMovementsSubscription.unsubscribe();
-    }
+    this.subscriptions.map( subscription => subscription.unsubscribe());
```
Drvanon: It is indeed a very good idea to keep this in it's own function, even though it's a one-liner, because this way it is much clearer what it is supposed to be doing.

Drvanon: The spacing in the map function call is inconsistent, choose either `call( argument );` or `call(argument);`, but not `call( argument);`.

gridt/src/app/login/login.page.ts
```diff
@@ -26,9 +26,7 @@ export class LoginPage implements OnDestroy {
   }
 
   private cancelAllSubscriptions() {
-    if (this.loginSubscription) {
-      this.loginSubscription.unsubscribe();
-    }
+    this.subscriptions.map( subscription => subscription.unsubscribe());
```
Drvanon: Spacing as mentioned before.

Drvanon: Maybe we could make this an interface/class? It is being used rather quite a lot.

gridt/src/app/login/register/register.page.ts
```diff
@@ -31,9 +31,7 @@ export class RegisterPage implements OnInit, OnDestroy {
   }
 
   private cancelAllSubscriptions() {
-    if (this.registerSubscription) {
-      this.registerSubscription.unsubscribe();
-    }
+    this.subscriptions.map( subscription => subscription.unsubscribe());
```
Drvanon: Spacing.

gridt/src/app/movements/add-movement/add-movement.page.ts
```diff
@@ -80,17 +75,17 @@ export class AddMovementPage implements OnInit, OnDestroy {
 
     await el.present();
 
-    this.createMovementSubscription = this.api.createMovement$(this.form.value as Movement).subscribe(
+    this.subscriptions.push(this.api.createMovement$(this.form.value as Movement).subscribe(
       (message) => {
         el.dismiss();
         this.showMessage(message);
-        this.getMovementSubscription = this.api.getMovement$(this.form.value.name).subscribe();
+        this.subscriptions.push(this.api.getMovement$(this.form.value.name).subscribe());
```
Drvanon: This will be rewritten in the future, allowing for a much more elegant way to do this.

gridt/src/app/core/api.service.ts
```diff
@@ -4,19 +4,29 @@ import { HttpClient } from "@angular/common/http";
...
 @Injectable({
   providedIn: "root"
 })
 export class ApiService {
+  get allMovements$ (): Observable<Movement[]> {
+    return this._allMovements$.asObservable();
+  }
+  get subscriptions$ (): Observable<Movement[]> {
+    return this._subscriptions$.asObservable();
+  }
```
Drvanon: Why the relocation? Also please maintain the spacing.

gridt/src/app/core/api.service.ts
```diff
@@ -4,19 +4,29 @@ import { HttpClient } from "@angular/common/http";
...
+
+  constructor (private http: HttpClient, private auth: AuthService) { }
```
Drvanon, line 20: Usually this is placed at the top of the class.

gridt/src/app/core/api.service.ts
```diff
@@ -110,7 +123,7 @@ export class ApiService {
         this.replace_movement_in_bsubject(this._subscriptions$, movement);
         this.replace_movement_in_bsubject(this._allMovements$, movement);
       })
-    )
+    );
```
Drvanon, line 127: Good catch

gridt/src/app/core/api.service.ts
```diff
@@ -127,7 +140,10 @@ export class ApiService {
       catchError( this.handleBadAuth() ),
       tap((movements) => this._allMovements$.next(movements)),
       map( () => true)
-    ).subscribe();
+    ).subscribe({
+      error(error) {console.log(error);
+      }
```
Drvanon: The formatting here is off. I would recommend

```ts
.subscribe({
    error: e => console.log(e)
});
```
or
```ts
.subscribe(()=>{}, (e) => console.log(e));
```

gridt/src/app/core/api.service.ts
```diff
@@ -136,14 +152,17 @@ export class ApiService {
...
         `${this.URL}/movements/subscriptions`,
         options
       )),
       tap( (movements) => this._subscriptions$.next(movements)),
       catchError( this.handleBadAuth() )
-    ).subscribe();
+    ).subscribe({
+      error(error) {console.log(error);
+      }
+    });
```
Drvanon: Indentation inconsistent.

Drvanon: Same comment on formatting.

gridt/src/app/core/models/subscriptions.model.ts
```diff
@@ -0,0 +1,9 @@
+import { Subscription } from 'rxjs';
+
+export class Subscriptions {
```
Drvanon: I think the naming of this class could use some work. What do you think of SubscriptionHolder?

Drvanon (changes requested):

gridt/src/app/core/api.service.ts
HendrikPreuter: This was a linter thing i am affraid. I have placed them back and it doesn't give me the same warning anymore so i am not exactly sure why it was that it wanted it back in front.

gridt/src/app/core/api.service.ts
```diff
@@ -230,17 +249,6 @@ export class ApiService {
...
-  * Observable to obtain identity from server.
-  */
-  public userIdentity$: Observable<Identity> = this.auth.readyAuthentication$.pipe(
-    flatMap((options) => this.http.get<Identity>(
-      `${this.URL}/identity`,
-      options
-    )),
-    catchError( this.handleBadAuth())
-  );
-
```
HendrikPreuter, line 243: this was changed due to my linter giving this:
Declaration of instance field not allowed after declaration of instance method. Instead, this should come at the beginning of the class/interface. (member-ordering)tslint(1)

gridt/src/app/core/models/subscriptions.model.ts
HendrikPreuter: I thought just subscriptions would be clear enough, since it is just the subscriptions. However i am also fine with the SubscriptionHolder name since that also acuratly describes what it does.

Drvanon:
Please make sure the checks pass before requesting a review. Also if you are still working on adding work to a pull request, please turn the pull request into a draft so the CI is not run all the time.

HendrikPreuter:
> Please make sure the checks pass before requesting a review. Also if you are still working on adding work to a pull request, please turn the pull request into a draft so the CI is not run all the time.

i will, that's why i have not re-requested feedback for now. I had an issue with merging again. Still trying to untangle that mess.

gridt/src/app/login/login.page.spec.ts
```diff
@@ -34,4 +35,10 @@ describe('LoginPage', () => {
   it('should create', () => {
     expect(component).toBeTruthy();
   });
+
+  it("should unsubscribe all subscriptions when leaving the page", () => {
+    component['loginSubscription'] = of(true).subscribe();
+    component.ngOnDestroy();
```
HendrikPreuter, line 46: test if subscriptions are added to subscriptions[]
test if subscriptions are unsubbed on destroy
spy on extendable class

gridt/src/app/core/models/subscription-holder.model.ts
```diff
@@ -0,0 +1,9 @@
+import { Subscription } from 'rxjs';
+
+export class SubscriptionHolder {
+    protected subscriptions: Subscription[] = [];
+
+    cancelAllSubscriptions () {
+        this.subscriptions.map( subscription => subscription.unsubscribe() );
+    }
+}
```
HendrikPreuter: add tests

HendrikPreuter: change name to subscription-holder.?.ts

HendrikPreuter: change location to core instead of models

PR #175, opened 2020-09-27
Identity service

### LINKED ISSUE #149: As a Client, I want a settings service which combines the server and local identity, so the front end is services with a single observable.
The user identity is stored both locally and on the server. A new service should be created that is able to retrieve both identities and combines them into a single observable.

requirements:
- If both local and server identity are available, the server identity should get priority;
- If the server identity is different from the local identity, overwrite the local identity with the server identity;
- When the server is unavailable only the local settings should be streamed.

marickmanrho:
Fixes #149 

Within the settings service a userIdentity obervable is added which retrieves the Identity of the user from the server or from localstorage.

At the moment only one test is failing due to not implementing the jasmine timer correctly. This I'll fix shortly. But at this stage already a review can be useful.

### DIFF
--- a/gridt/src/app/core/api.service.spec.ts
+++ b/gridt/src/app/core/api.service.spec.ts
@@ -41,7 +41,7 @@ let mock_movements: Movement[] = [
     id: 112312983,
     name: "Flossing",
     short_description: "Floss every day",
-    description: "We floss every day because it is good for our theeth.",
+    description: "We floss every day because it is good for our teeth.",
     subscribed: true,
     interval: "daily"
   }
@@ -243,7 +243,7 @@ describe("ApiService when authentication is succesful", () => {
     const flossing_movement = {
       name: "Flossing",
       short_description: "Floss every day",
-      description: "We floss every day because it is good for our theeth.",
+      description: "We floss every day because it is good for our teeth.",
       subscribed: true,
       interval: "daily"
     } as Movement;
--- a/gridt/src/app/core/api.service.spec.ts
+++ b/gridt/src/app/core/api.service.spec.ts
@@ -129,7 +129,7 @@ describe("ApiService when authentication fails", () => {
 
   it('should fail to read identity from server when not logged in', () => {
     expect(
-      service.userIdentity$
+      service.userIdentity$()
     ).toBeObservable(cold("#", {}, "Can't authenticate: no credentials"));
   });
 
@@ -312,8 +312,8 @@ describe("ApiService when authentication is succesful", () => {
 
   it('should be able to retreive identity', () => {
     httpClientStub.get.and.returnValue(of(mock_identity[0]));
-    
-    expect(service.userIdentity$).toBeObservable(cold('(a|)', {a: mock_identity[0]}));
+
+    expect(service.userIdentity$()).toBeObservable(cold('(a|)', {a: mock_identity[0]}));
     expect(httpClientStub.get).toHaveBeenCalledWith(
       `${service.URL}/identity`,
       default_headers
--- a/gridt/src/app/core/api.service.spec.ts
+++ b/gridt/src/app/core/api.service.spec.ts
@@ -1,7 +1,6 @@
 import { Type } from "@angular/core";
 import { TestBed } from "@angular/core/testing";
 import { HttpClient, HttpHeaders, HttpErrorResponse } from '@angular/common/http';
-import { HttpClientTestingModule, HttpTestingController } from "@angular/common/http/testing";
 
 import { cold } from 'jasmine-marbles';
 import { of, throwError } from "rxjs";
@@ -81,10 +80,7 @@ const default_headers = {
 };
 
 let service: ApiService;
-let auth: AuthService;
-let httpMock: HttpTestingController;
 let httpClientStub: jasmine.SpyObj<HttpClient>;
-let authServiceStub: jasmine.SpyObj<AuthService>;
 
 class authServiceStub_succes {
   isLoggedIn$ = of(true);
@@ -128,14 +124,14 @@ describe("ApiService when authentication fails", () => {
   });
 
   it('should fail to read identity from server when not logged in', () => {
-    expect(
-      service.userIdentity$()
-    ).toBeObservable(cold("#", {}, "Can't authenticate: no credentials"));
+    expect(service.userIdentity$())
+    .toBeObservable(cold("#", {}, "Can't authenticate: no credentials"));
   });
 
   it('should fail to update bio when not logged in', () => {
     let mock_bio = "My very first bio."
-    expect(service.changeBio$( mock_bio )).toBeObservable( cold('#', null, "Can't authenticate: no credentials"));
+    expect(service.changeBio$( mock_bio ))
+    .toBeObservable( cold('#', null, "Can't authenticate: no credentials"));
   });
 
   it('should fail to update password when not logged in', () => {
--- a/gridt/src/app/core/settings.service.spec.ts
+++ b/gridt/src/app/core/settings.service.spec.ts
@@ -99,7 +99,7 @@ describe("SettingsService when authentication fails", () => {
 
   it("should return an empty identity", () => {
     // spyOn(api, 'userIdentity$').and.returnValue(of(mock_id[0]));
-    apiStub.userIdentity$.and.returnValue( of(mock_id[0]) );
+    apiStub.userIdentity$.and.returnValue(of(mock_id[0]));
     service.updateIdentity();
     expect(service.userIdentity$).toBeObservable(
       cold('a', {a: service.empty_identity})
@@ -123,14 +123,16 @@ describe("SettingsService when authentication fails", () => {
 
 describe("SettingsService when authentication is succesful", () => {
   beforeEach( () => {
+    jasmine.clock().install();
+
     apiStub = jasmine.createSpyObj('api', {'userIdentity$': mock_id[0]});
 
     secStoreStub = jasmine.createSpyObj(
       'secStore', {
-        'get$': of(JSON.stringify(mock_id[0])),
-        'set$': of(true),
-        'clear$': of(true),
-        'remove$': of(true)
+        'get$'    : of(JSON.stringify(mock_id[0])),
+        'set$'    : of(true),
+        'clear$'  : of(true),
+        'remove$' : of(true)
       }
     );
 
@@ -151,14 +153,11 @@ describe("SettingsService when authentication is succesful", () => {
     service = TestBed.get(SettingsService as Type<SettingsService>);
     api = TestBed.get(ApiService as Type<ApiService>);
 
-    let tick: (milliseconds: number) => void;
-    let fakeNow = 0;
-    tick = milliseconds => {
-      fakeNow += milliseconds;
-      tick(milliseconds);
-    };
-    asyncScheduler.now = () => fakeNow;
   });
+
+  afterEach( () => {
+    jasmine.clock().uninstall();
+  })
 
   it("should be created", () => {
     expect(service).toBeTruthy();
@@ -166,14 +165,14 @@ describe("SettingsService when authentication is succesful", () => {
 
   it("should return the local id when server id is unavailable", () => {
     // When there is no internet, the identity must remain available.
-    apiStub.userIdentity$.and.returnValue( throwError("server unavailable") );
+    apiStub.userIdentity$.and.returnValue(throwError("server unavailable"));
     service.updateIdentity();
     expect(service.userIdentity$).toBeObservable(cold('a',{a:mock_id[0]}));
   });
 
   it("should overwrite local id when server id differs", () => {
     // Make sure any updates of the ID are reflected in the local storage.
-    apiStub.userIdentity$.and.returnValue( of(mock_id[1]) );
+    apiStub.userIdentity$.and.returnValue(of(mock_id[1]));
     service.updateIdentity();
     expect(secStoreStub.set$).toHaveBeenCalledWith(
       "identity", JSON.stringify(mock_id[1])
@@ -182,7 +181,7 @@ describe("SettingsService when authentication is succesful", () => {
 
   it("should not overwrite local id when local and server id are equal", () => {
     // Reduces number of calls
-    apiStub.userIdentity$.and.returnValue( of(mock_id[0]) );
+    apiStub.userIdentity$.and.returnValue(of(mock_id[0]));
     service.updateIdentity();
     expect(secStoreStub.set$).not.toHaveBeenCalled();
   });
@@ -190,7 +189,7 @@ describe("SettingsService when authentication is succesful", () => {
   it("should set local id with server id when local id is unavailable", () => {
     // This happens for new users, because their localstorage has not been set yet.
     secStoreStub.get$.and.returnValue(throwError("Oops, localstore not set yet"));
-    apiStub.userIdentity$.and.returnValue( of(mock_id[3]));
+    apiStub.userIdentity$.and.returnValue(of(mock_id[3]));
     service.updateIdentity();
     expect(secStoreStub.set$).toHaveBeenCalledWith(
       "identity", JSON.stringify(mock_id[3])
@@ -200,7 +199,7 @@ describe("SettingsService when authentication is succesful", () => {
   it("should return empty identity when local and server are unavailable", () => {
     // Safety feature to make sure it doesn't prevent a page from loading.
     secStoreStub.get$.and.returnValue(throwError("Oops, localstore not set yet"));
-    apiStub.userIdentity$.and.returnValue( throwError("Server not available"));
+    apiStub.userIdentity$.and.returnValue(throwError("Server not available"));
     service.updateIdentity();
     expect(service.userIdentity$).toBeObservable(
       cold('a',{a:service.empty_identity})
@@ -210,7 +209,7 @@ describe("SettingsService when authentication is succesful", () => {
   it("should return local id when server and local id are equal", () => {
     //  Standard behaviour
     secStoreStub.get$.and.returnValue(of(JSON.stringify(mock_id[0])));
-    apiStub.userIdentity$.and.returnValue( of(mock_id[0]) );
+    apiStub.userIdentity$.and.returnValue(of(mock_id[0]));
     service.updateIdentity();
     expect(service.userIdentity$).toBeObservable(cold('a',{a: mock_id[0]}));
   });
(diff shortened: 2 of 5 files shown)


marickmanrho:
The observable `api.userIdentity$` has been made into a function because Jasmine didn't like the fact that is was not a function and thus prevented me to stub it properly.

Drvanon:
In general this problem of not being able to stub should be able to be fixed by stubbing the class. I haven't looked into the code yet, but could that be a solution?

gridt/src/app/core/settings.service.spec.ts
```diff
@@ -93,32 +114,163 @@ describe("SettingsService when authentication fails", () => {
...
-  })
+    api = TestBed.get(ApiService as Type<ApiService>);
+
+    let tick: (milliseconds: number) => void;
+    let fakeNow = 0;
+    tick = milliseconds => {
+      fakeNow += milliseconds;
+      tick(milliseconds);
+    };
+    asyncScheduler.now = () => fakeNow;
```
Drvanon: In home.page.spec.ts or the swap.service.spec.ts file presents a way of setting the time in a more simple way if I understand the purpose of this piece of code correctly. If not, could you comment as to why you used this?

gridt/src/app/core/settings.service.spec.ts
```diff
@@ -93,32 +114,163 @@ describe("SettingsService when authentication fails", () => {
...
+    // Reduces number of calls
+    apiStub.userIdentity$.and.returnValue( of(mock_id[0]) );
+    service.updateIdentity();
+    expect(secStoreStub.set$).not.toHaveBeenCalled();
+  });
+
+  it("should set local id with server id when local id is unavailable", () => {
+    // This happens for new users, because their localstorage has not been set yet.
+    secStoreStub.get$.and.returnValue(throwError("Oops, localstore not set yet"));
+    apiStub.userIdentity$.and.returnValue( of(mock_id[3]));
```
Drvanon: We should create a js lint rule for the spacing in arguments, it's silly that this should be caught by humans.

Drvanon: Also I notice that you/we are not consistent with this spacing. It would be good to come to a decision and run a formatter to make the entire repository adopt the rule.

gridt/src/app/core/settings.service.spec.ts
```diff
@@ -93,32 +114,163 @@ describe("SettingsService when authentication fails", () => {
...
+
+  it("should return local id when server and local id are equal", () => {
+    //  Standard behaviour
+    secStoreStub.get$.and.returnValue(of(mock_id[0]));
+    apiStub.userIdentity$.and.returnValue( of(mock_id[0]) );
+    service.updateIdentity();
+    expect(service.userIdentity$).toBeObservable(cold('a',{a: mock_id[0]}));
+  });
+
+  it("should deal with many async calls", () => {
```
Drvanon, line 209: Color me impressed.

gridt/src/app/core/settings.service.ts
```diff
@@ -22,26 +24,97 @@ export class SettingsService {
...
+  private mapErrorToEmptyID = pipe(
+    catchError( err => {console.warn(err); return of<Identity>(this.empty_identity);}),
+    flatMap( (id) => of(<Identity>(id)))
+  );
+
+  private mapErrorToFalse = pipe(
+    catchError( err => {console.warn(err); return of<boolean>(false);})
+  );
+
+  public updateIdentity(): void {
```
Drvanon, line 78: I'm on my phone right now but first impression is that this is really well thought out and I cant find any failures. Also for something this complicated it is super great to have this many comments.

Drvanon (commented):
Super nice commit, will have to look into this a bit deeper, on a computer.

gridt/src/app/core/settings.service.spec.ts
marickmanrho: I just looked around the internet for a solution. Will be happy to use a neater method.

Drvanon: https://github.com/GridtNetwork/gridt-client/blob/5d1642df642846e64ccb7ae58e900ee62dc3158c/gridt/src/app/home/home.page.spec.ts#L24 

Does this what you want it to do?

marickmanrho: Well... I've tried that but it didn't work on the rxjs timer.

marickmanrho: Okey, it works now! Thanks for pointing me to the home.page.spec.ts example

gridt/src/app/core/settings.service.spec.ts
```diff
@@ -244,6 +265,10 @@ describe("SettingsService when authentication is succesful", () => {
   });
 
   it("should wait for the server to respond", fakeAsync(() => {
+    // let fake_now = new Date();
+    // fake_now.setUTCHours(1,0,0,0);
```
Drvanon: Why is this commented out? Should it be removed or is it a WIP

gridt/src/app/core/api.service.spec.ts
```diff
@@ -129,7 +129,7 @@ describe("ApiService when authentication fails", () => {
 
   it('should fail to read identity from server when not logged in', () => {
     expect(
-      service.userIdentity$
+      service.userIdentity$()
     ).toBeObservable(cold("#", {}, "Can't authenticate: no credentials"));
```
HendrikPreuter: The style of these lines of code seem in contrast with the others as they use:
`expect(service.changeBio$( mock_bio )).toBeObservable( cold('#', null, "Can't authenticate: no credentials"));`

If you do want to split them up into different lines i think it should be like this:

`   expect(service.userIdentity$())`
`.toBeObservable(cold("#", {}, "Can't authenticate: no credentials"));`

gridt/src/app/core/settings.service.spec.ts
```diff
@@ -1,21 +1,26 @@
 import { Type } from "@angular/core";
-import { TestBed } from "@angular/core/testing";
+import { TestBed, fakeAsync, tick } from "@angular/core/testing";
 import { HttpClient, HttpHeaders, HttpErrorResponse } from '@angular/common/http';
 
 import { cold } from 'jasmine-marbles';
-import { of, throwError } from "rxjs";
+import { of, throwError, timer, asyncScheduler } from "rxjs";
+import { take, flatMap } from "rxjs/operators"
```
HendrikPreuter: missing a ;

gridt/src/app/core/settings.service.spec.ts
```diff
@@ -1,21 +1,26 @@
 import { Type } from "@angular/core";
-import { TestBed } from "@angular/core/testing";
+import { TestBed, fakeAsync, tick } from "@angular/core/testing";
 import { HttpClient, HttpHeaders, HttpErrorResponse } from '@angular/common/http';
 
 import { cold } from 'jasmine-marbles';
-import { of, throwError } from "rxjs";
+import { of, throwError, timer, asyncScheduler } from "rxjs";
```
HendrikPreuter: HttpErrorResponse and asyncSchedular seem like they are never used

gridt/src/app/core/settings.service.spec.ts
```diff
@@ -1,21 +1,26 @@
...
 import { SettingsService } from "./settings.service";
 import { SecureStorageService } from './secure-storage.service';
 
 import { Identity } from "./models/identity.model";
 
 let service: SettingsService;
 let auth: AuthService;
+let api: ApiService;
 let secStore: SecureStorageService;
 let authServiceStub: jasmine.SpyObj<AuthService>;
```
HendrikPreuter: auth, secStore and authServiceStub are never used

gridt/src/app/core/settings.service.spec.ts
HendrikPreuter: HttpErrorResponse and asyncSchedular are never used

gridt/src/app/core/settings.service.spec.ts
HendrikPreuter: Missing a ;

gridt/src/app/core/settings.service.spec.ts
HendrikPreuter: HttpErrorResponse and asyncSchedular are never used

gridt/src/app/core/settings.service.spec.ts
HendrikPreuter: auth, secStore and authServiceStub are never used

gridt/src/app/core/api.service.spec.ts
HendrikPreuter: Styling seems to clash with the other expects which are just one line.
If you do want this on other lines i think you should split them like this:

`    expect(
      service.userIdentity$()
    )`
`.toBeObservable(cold("#", {}, "Can't authenticate: no credentials"));`

gridt/src/app/core/settings.service.ts
```diff
@@ -22,27 +24,98 @@ export class SettingsService {
...
+    // Combine server and local id. Catch errors to make sure pipe runs.
+    forkJoin({
+      server: this.api.userIdentity$().pipe( this.mapErrorToEmptyID ),
+      local: this.localIdentity$.pipe( this.mapErrorToEmptyID )
+    }).pipe(
+      // Update local id when server id is different.
+      tap( (ids) => console.debug(`recieved server id = ${JSON.stringify(ids.server)} \n received local id = ${JSON.stringify(ids.local)}`)),
+      flatMap( (ids) => iif(
+        () => ((JSON.stringify(ids.server)==JSON.stringify(ids.local)) || (JSON.stringify(ids.server) == JSON.stringify(this.empty_identity))), // && ids[0].id == ids[1].id
```
HendrikPreuter: should these not be triple =? and check for spacing around the ==

gridt/src/app/core/settings.service.ts
```diff
@@ -22,27 +24,98 @@ export class SettingsService {
...
+        of(true),
+        this.setLocalIdentity$(ids.server)
+      )),
+      // Make sure any errors arising from setting local identity are
+      // transformed into a warning and continue.
+      this.mapErrorToFalse,
+      // Emit local identity
+      flatMap( () => this.localIdentity$ ),
+      // If local identity unvailable emit empty ID.
```
HendrikPreuter: unavailable

gridt/src/app/core/settings.service.ts
```diff
@@ -22,27 +24,98 @@ export class SettingsService {
...
+  /**
+  * This Observable can be used in applications to display the latest known
+  * identity of the user.
+  */
+  public userIdentity$: Observable<Identity> = this._user_identity$.asObservable();
+
+  /**
+  * This function updates the user identity.
+  * - If both server and local identity are unavailable, emit empty ID.
+  * - When server ID is retreived, but local ID has not been set yet, store
```
HendrikPreuter: retrieved

gridt/src/app/core/settings.service.ts
HendrikPreuter: line 81 is also quite long, i would suggest splitting it into a new line after the ||

gridt/src/app/core/settings.service.ts
```diff
@@ -22,27 +24,98 @@ export class SettingsService {
...
+   * @param identity Identity to be stored.
    */
   public setLocalIdentity$(identity: Identity): Observable<boolean> {
     return this.auth.readyAuthentication$.pipe(
       catchError( () => throwError(this.error_codes.SETIDFAIL + ": " + this.error_codes.NOTLOGGEDIN)),
       take(1), // Makes sure the observable completes
-      flatMap( () => this.secStore.set$("identity", identity).pipe(
+      tap( () => console.debug('Storing Identity in Localstorage')),
+      flatMap( () => this.secStore.set$("identity", JSON.stringify(identity)).pipe(
           catchError(()=> throwError(this.error_codes.SETIDFAIL + ": " + this.error_codes.SECSTOREUNAVAILABLE))
```
HendrikPreuter: spacing catchError( () => etc.)

HendrikPreuter (commented):
Seems promising. I found only a few style improvements. Good work!

gridt/src/app/core/settings.service.spec.ts
```diff
@@ -93,32 +123,183 @@ describe("SettingsService when authentication fails", () => {
...
+    secStoreStub.set$.and.callFake( (key: string, value: any) => {
+      secStoreStub.get$.and.returnValue(of(value));
+      return of(true);
+    });
+
+    service.updateIdentity();
+    tick(15);
+    expect(service.userIdentity$).toBeObservable(cold('',{}));
+    tick(30);
+    expect(service.userIdentity$).toBeObservable(cold('a',{a:mock_id[1]}));
```
Drvanon: As far as I understand, changing the local time does not have an effect on the observables, as time is being mocked in it's entirety by jasmine-marbles and the executing of it takes place instantly.

gridt/src/app/core/settings.service.ts
Drvanon: Also: you leave commented code at the end of this line. Please remove it before merger.

gridt/src/app/core/settings.service.ts
```diff
@@ -22,27 +24,98 @@ export class SettingsService {
...
+   * Observable to obtain identity from secure storage
    * If secure storage is empty, this returns an unavailable error
    */
   public localIdentity$: Observable<Identity> = this.auth.readyAuthentication$.pipe(
     catchError( () => throwError(this.error_codes.GETIDFAIL + ": " + this.error_codes.NOTLOGGEDIN)),
     flatMap(() => this.secStore.get$("identity").pipe(
       catchError( () => throwError(this.error_codes.GETIDFAIL + ": " + this.error_codes.SECSTOREUNAVAILABLE))
-    ))
+    )),
+    flatMap( (id) => of<Identity>(JSON.parse(id)))
```
Drvanon: As far as I understand, the output of secstore is already parsed. Why are you parsing it again?

Drvanon (commented):
Looking good. Please take a look at the comments of @HendrikPreuter, I don't see any logical mistakes.

gridt/src/app/core/settings.service.spec.ts
marickmanrho: I'm not sure what you mean with this comment. Do you mean I could leave out
```
    tick(15);
    expect(service.userIdentity$).toBeObservable(cold('',{}));
    tick(30);
````
?

File: gridt/src/app/core/settings.service.ts (hunk 15 above)
Comment (marickmanrho) [reply gridt/src/app/core/settings.service.ts]: No?? I don't think so, here is a copy of the secStore.get$ function
```
get$ (key: string): Observable<any> {
    return new Observable ( (observer) => {
      SecureStoragePlugin.get({key}).then(
        (valueObj: any) => {
          const value = valueObj.value;
          if (value === atob(null) || !value){
            observer.error(`Key "${key}" does not exist in the secure storage.`);
            return;
          } else {
            observer.next(value);
            observer.complete();
          }
        }
      ).catch( () => observer.error(`Key "${key}" does not exist in the secure storage.`) );
    });
```
This simply outputs the value as was stored in the secStore.

File: gridt/src/app/core/settings.service.spec.ts (hunk 14 above)
Comment (Drvanon) [reply gridt/src/app/core/settings.service.spec.ts]: I mean that you could take out the `tick()`s.

File: gridt/src/app/core/settings.service.ts (hunk 15 above)
Comment (Drvanon) [reply gridt/src/app/core/settings.service.ts]: Oh dear, this is a problem, I have added that functionality some where, maybe it is still in an active pull request?

File: gridt/src/app/core/settings.service.spec.ts (hunk 14 above)
Comment (marickmanrho) [reply gridt/src/app/core/settings.service.spec.ts]: Yet if I do, the test fails. The idea of this test is to mock the server response with the timer function. This only fires after 30ms. Hence, up to that point the userIdentity$ must remain undefined. After 30ms it should be set. That timing must be done with the tick() functions. Good thing is that Jasmine takes care of the tick() functions and thus indeed evaluates everything at once.

Comment (Drvanon) [reply gridt/src/app/core/settings.service.spec.ts]: Then I probably lack a good understanding of your code. Let me take another look

File: gridt/src/app/core/settings.service.ts (hunk 15 above)
Comment (marickmanrho) [reply gridt/src/app/core/settings.service.ts]: Remember that time my draft pulled into development? I thought I fixed it by just reverting the changes. Turns out there where some changes made after that mistake. Guess what... That was your secStore change. So unfortunately those changes got overwritten... Not sure how to continue, should we accept this PR, open a new issue and include the changes you made earlier. Or shall I include your changes in this PR?

File: gridt/src/app/core/settings.service.ts [hunk 16]
```diff
@@ -22,28 +24,102 @@ export class SettingsService {
...
+  * - If both the server and local identity agree, emit local ID.
+  * - If the server is unavailable but local ID is available, emit local ID.
+  * - Only when updateIdentity() function is called should the userIdentity$
+  *   change.
+  *
+  * - Makes sure all errors are caught such that this function always gives an
+  *   output.
+  */
+  private mapErrorToEmptyID = pipe(
+    catchError( err => {console.warn(err); return of<Identity>(this.empty_identity);}),
```
HendrikPreuter: This looks somewhat weird to me.
i think it would be better to again split this into multiple different lines.

gridt/src/app/core/settings.service.ts
```diff
@@ -22,28 +24,102 @@ export class SettingsService {
...
+    flatMap( (id) => of(<Identity>(id)))
+  );
+
+  private mapErrorToFalse = pipe(
+    catchError( err => {console.warn(err); return of<boolean>(false);})
```
HendrikPreuter: same here

gridt/src/app/core/settings.service.ts
```diff
@@ -22,28 +24,102 @@ export class SettingsService {
...
+   * @param identity Identity to be stored.
    */
   public setLocalIdentity$(identity: Identity): Observable<boolean> {
     return this.auth.readyAuthentication$.pipe(
       catchError( () => throwError(this.error_codes.SETIDFAIL + ": " + this.error_codes.NOTLOGGEDIN)),
       take(1), // Makes sure the observable completes
+      tap( () => console.debug('Storing Identity in Localstorage')),
       flatMap( () => this.secStore.set$("identity", identity).pipe(
-          catchError(()=> throwError(this.error_codes.SETIDFAIL + ": " + this.error_codes.SECSTOREUNAVAILABLE))
+          catchError( ()=> throwError(this.error_codes.SETIDFAIL + ": " + this.error_codes.SECSTOREUNAVAILABLE))
```
HendrikPreuter: here too

gridt/src/app/core/settings.service.ts
```diff
@@ -22,28 +24,102 @@ export class SettingsService {
...
-   * Store settings in the secure storage.
-   * @param identity Settings to be stored.
+   * Store identity in the secure storage.
+   * @param identity Identity to be stored.
    */
   public setLocalIdentity$(identity: Identity): Observable<boolean> {
     return this.auth.readyAuthentication$.pipe(
       catchError( () => throwError(this.error_codes.SETIDFAIL + ": " + this.error_codes.NOTLOGGEDIN)),
       take(1), // Makes sure the observable completes
+      tap( () => console.debug('Storing Identity in Localstorage')),
```
HendrikPreuter: don't forget the space at the end there aswell!
either (statement) or ( statement ) and not the ( statement)
I have made that mistake too.

HendrikPreuter (commented):
Api.service.spec.ts still has theeth in it on lines 44 and 246, this should be teeth instead.

Drvanon (approved):
I think you incorporated all the feedback I had. Let's merge!

PR #178, opened 2020-10-02
Re-apply commit - Make secure storage store objects as well.

marickmanrho:
Due to an error these changes have been overwritten. Now they are reinstated.


Drvanon:
Welcome back to the code base.

PR #183, opened 2020-10-08
Add location & categories & tests to add movement

### LINKED ISSUE #166: As a user, i want to be able to set a category and location attribute when making movements
#165 
Up: 1
Upr: 3

To-do: change the front-end to force the user to choose a catagory for the movement

Files affected:
Createmovent.html/scss
Possibly ".js to communicate with the new/updated api

HendrikPreuter:
#166 

Done:
- tests to indicate whether or not the button is disabled
- added categories and location
- added compulsory statement

### DIFF
--- a/gridt/src/app/movements/add-movement/add-movement.page.html
+++ b/gridt/src/app/movements/add-movement/add-movement.page.html
@@ -94,7 +94,7 @@
         </ion-col>
       </ion-row>
       <ion-row>
-        <ion-col>
+        <ion-col size-sm="6" offset-sm="3">
           <small>Items marked with * are compulsory</small>
         </ion-col>
       </ion-row>
--- a/gridt/src/app/movements/add-movement/add-movement.page.html
+++ b/gridt/src/app/movements/add-movement/add-movement.page.html
@@ -5,7 +5,7 @@
     </ion-buttons>
     <ion-title>New Movement</ion-title>
     <ion-buttons slot="primary">
-      <ion-button (click)="confirmCreation()" [disabled]="!form.valid">
+      <ion-button id="confirmCreation" (click)="confirmCreation()" [disabled]="!form.valid">
         <ion-icon name="checkmark" slot="icon-only"></ion-icon>
       </ion-button>
     </ion-buttons>
@@ -18,7 +18,7 @@
       <ion-row>
         <ion-col size-sm="6" offset-sm="3">
           <ion-item>
-            <ion-label position="floating">Movement name</ion-label>
+            <ion-label position="floating">Movement name*</ion-label>
             <ion-input type="text" autocomplete autocorrect ngDefaultControl formControlName="name"></ion-input>
           </ion-item>
         </ion-col>
@@ -35,7 +35,7 @@
       <ion-row>
         <ion-col size-sm="6" offset-sm="3">
           <ion-item>
-            <ion-label position="floating">Short Description</ion-label>
+            <ion-label position="floating">Short Description*</ion-label>
             <ion-textarea rows="2" ngDefaultControl formControlName="short_description"placeholder="What are we going to do?"></ion-textarea>
           </ion-item>
         </ion-col>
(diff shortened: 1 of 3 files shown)


gridt/src/app/movements/add-movement/add-movement.page.html
```diff
@@ -18,7 +18,7 @@
       <ion-row>
         <ion-col size-sm="6" offset-sm="3">
           <ion-item>
-            <ion-label position="floating">Movement name</ion-label>
+            <ion-label position="floating">Movement name*</ion-label>
```
marickmanrho, line 21: Why the asterisk?

gridt/src/app/movements/add-movement/add-movement.page.spec.ts
```diff
@@ -20,11 +20,11 @@ class AuthServiceStub {
   readyAuthentication$ = of(default_headers)
 }
 
-describe('AddMovementPage', () => {
+describe("AddMovementPage", () => {
   let component: AddMovementPage;
   let fixture: ComponentFixture<AddMovementPage>;
   let authSpy: AuthService;
-  let alertSpy: AlertController = jasmine.createSpyObj("alertSpy", ["create", "dismiss"]);
+  const alertSpy: AlertController = jasmine.createSpyObj("alertSpy", ["create", "dismiss"]);
```
marickmanrho: Why is it important that it's a const?

gridt/src/app/movements/add-movement/add-movement.page.spec.ts
```diff
@@ -48,12 +48,131 @@ describe('AddMovementPage', () => {
...
+    compiled.form.patchValue({
+      name: "test",
+      short_description: "this is a short description",
+      description: "this is a longer description that the short description to be able to tell the difference",
+      interval: "daily",
+      category: "Community",
+      location: "Groningen"
+    });
+    fixture.detectChanges();
+    expect(submitEl.disabled).not.toBeTruthy();
```
marickmanrho: `submitEl.disabled` should actually be false right? Not just Falsy, which is a broader statement I think.

marickmanrho: See also [this question on stackoverflow](https://stackoverflow.com/questions/32615713/tobetrue-vs-tobetruthy-vs-tobetrue)

gridt/src/app/movements/add-movement/add-movement.page.spec.ts
```diff
@@ -48,12 +48,131 @@ describe('AddMovementPage', () => {
...
+    const submitEl = fixture.debugElement.nativeElement.querySelector("#confirmCreation");
+    expect(submitEl.disabled).toBeTruthy();
+  });
+
+  it("should disable the confirm button when the name is filled in incorrectly", () => {
+    const submitEl = fixture.debugElement.nativeElement.querySelector("#confirmCreation");
+    const compiled = fixture.componentInstance;
+    compiled.form.patchValue({
+      short_description: "this is a short description",
+      description: "this is a longer description that the short description to be able to tell the difference",
```
marickmanrho: than

gridt/src/app/movements/add-movement/add-movement.page.ts
```diff
@@ -17,6 +17,27 @@ export class AddMovementPage implements OnInit, OnDestroy {
...
+    'Noord-Brabant',
+    'Zeeland'
+  ];
+  categories: string[] = [
+    'Health & Fitness',
+    'Mind & Learning',
+    'Community',
+    "Sustainable Planet",
+    "Miscellaneous"
+  ];
```
marickmanrho, line 40: Good fix for now. I think it is good if everyone is aware of this such that we can move these lists to the server one day.

marickmanrho (changes requested):
When tests are sorted we can merge.

gridt/src/app/movements/add-movement/add-movement.page.html
HendrikPreuter: Just some extra user feedback to let them know that is a mandatory field, even before entering anything.

gridt/src/app/movements/add-movement/add-movement.page.spec.ts
HendrikPreuter: My linter keeps telling me that it should be constants because they never get changed. So i just accepted it's requests this time. However i think we might need some general consensus on these thngs so that the code stays uniform. Either all Lets or const if they don;t get reassigned.

gridt/src/app/movements/add-movement/add-movement.page.spec.ts
HendrikPreuter: thank you!

gridt/src/app/movements/add-movement/add-movement.page.spec.ts
HendrikPreuter: I will look into this

HendrikPreuter: That seems to be the correct way of doing this, so i will change it.

marickmanrho (approved):

PR #184, opened 2020-10-20
Android clean up

AndraUdrescu:
Deleted everything we don't need from android. This files will be created again if we make an android build. Second branch  because I accidentally made changes on a very old branch. That branch needs to be deleted. It is clean-up-android


Drvanon:
This does not remove the .jks file. Also, please include the android folder in the .gitignore.

Drvanon:
Quick feedback on git commit naming conventions: we generally write commit messages in the imperative mood. For this last commit (bd45b90) I would suggest writing: "Remove .jsk file and add android files to .gitignore". You can leave it for now, but just as a heads up.

PR #185, opened 2020-10-29
165 category api

### LINKED ISSUE #165: As a client, i want to be able to set a category and location attribute for movements
Up: 2
Upr: 6

To-do:
Write or change an api that forces the user to choose a catagory for the movement they are creating.

Files affected:

movement.model.ts

HendrikPreuter:
#165 
Done:
- changed movement.model,ts
- add 1 test

### DIFF
--- a/gridt/src/app/core/models/movement.model.ts
+++ b/gridt/src/app/core/models/movement.model.ts
@@ -44,7 +44,7 @@ export interface Movement {
   location?: string;
 
     /**
-   * A number of catagories between 1-5 to help sort the movements.
+   * A number of categories between 1-5 to help sort the movements.
    */
   category?: Array<string>;
 
--- a/gridt/src/app/core/models/movement.model.ts
+++ b/gridt/src/app/core/models/movement.model.ts
@@ -38,5 +38,15 @@ export interface Movement {
    */
   description?: string;
 
+    /**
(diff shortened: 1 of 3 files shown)


gridt/src/app/core/models/movement.model.ts
```diff
@@ -38,5 +38,15 @@ export interface Movement {
    */
   description?: string;
 
+    /**
+   * A location where the movement is active.
+   */
+  location?: string;
```
marickmanrho, line 44: Could the location be an Array also? I can imagine a movement being in the north of the Netherlands for example.

marickmanrho (approved):
Great, only small question.

gridt/src/app/core/models/movement.model.ts
HendrikPreuter: It can become that later on, however for this initial request we agreed on just the provinces. Now from my point of view that meant 1 province. But that can change as a separate issue is made for the location.

giourikilin:
@AntoninThioux and I want to know if this is still relevant or not needed any more @Yori-O92 ?
Backend has no implementation regarding this changes.

Yori-O92:
@AntoninThioux @giourikilin Let's not bother ourselves with this for now.

PR #196, opened 2023-04-17
192 see announcements

### LINKED ISSUE #192: As a user I want to be able to see announcements for movements
There should be a way to see announcements made by the movement administrators for each movement a user is subscribed to.

AntoninThioux:
Resolves #192 regarding letting users view announcements made in the movement.


giourikilin (approved):
Nice

Issue #193, opened 2023-04-03
As a administrator I want to be able to create announcements for movements


AntoninThioux:
There should be a UI for the administrators to use the new backend endpoints to create new movement announcements.

PR #197, opened 2023-04-18
193 create announcements

AntoninThioux:
Implements the create announcement feature described by the user story in #193. Additionally, I added buttons to update and remove announcements.


gridt/src/app/controller/announcement-controller/announcement.page.ts
giourikilin, line 1: It is nice, but I think the checks for empty "message" to avoid empty announcements, you should also check for empty space character, since if you press "space bar" 11 times the "message.length" will be 11 and an empty announcement will passed.

giourikilin (changes requested):
Can you test locally if by creating an announcement message and pressing "space bar" 11 times allow you to create an empty announcement

gridt/src/app/controller/announcement-controller/announcement.page.ts
AntoninThioux: Good point I pushed a fix!

giourikilin (approved):

giourikilin:
Awesome!

PR #201, opened 2023-11-15
Lay-out changes for the ppc symposium

Yori-O92:
Hi Antonin,

I made a few textual and visual changes in the app prototype for the pilot at the ppc symposium. I'm not able to test on my machine (I should set it up again), but I think nothing breaks this way. I hope I didn't mess up anything. Could you review and merge if approved?


AntoninThioux (approved):
I tested it on my machine everything works

